PR #46, opened 2026-07-02
v0.10.2 [SPEC]: §6.6 owed discharge requires the owing sender (N1, spec-first)

AlexAvilov:
PR-A-spec (split from PR-A per CONTRIBUTING.md §Spec-changes — spec-only PR first, code follows). Authored by codex.

**Change:** §6.6 — the asker's `.owed` obligation is discharged on a consumed reply only if the reply's **canonical sender is the agent that owed it**. Previously the sentence was sender-agnostic.

**Why spec-first (separate PR):** §6.6 is a listed covenant (spec line 3) and CONTRIBUTING.md:74-82 requires spec changes to land as their own PR before the code. Caught by sonnet in plan review; verified against CONTRIBUTING.md.

**Patch-safe:** `.owed` is asker-local, non-wire, non-blocking (a warning, never a finish blocker). This tightens *when a local ledger clears* — no envelope/filename/interop/conformance change. codex ratified the covenant call.

**Merge order:** this PR first; the code guard (PR-A-code, #next) merges after.

Gates: codex (author — needs sonnet + claude) → so **sonnet + claude** gate.

### DIFF
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -184,7 +184,7 @@ Receivers MUST ignore unrecognized frontmatter fields. `from` is required inform
 Reply obligations are **asker-owned only**. The asker's `.owed` ledger is the **sole** reply-obligation mechanism (non-blocking warning + expiry). **Recipients are never blocked at finish by a `reply_required` message** — delivery is best-effort pull, with no forcing function once delivered.
 
 - When an agent sends `--ask` (reply-required), it records its own obligation in `state/<asker>/owed.json`: "I am owed a reply to `(to=recipient, from=me, seq)` by `<deadline>`" (default 30m; override with `--reply-by`). The ledger is single-writer, atomic-rename, and the gate reads only its OWN ledger — it never scans peers.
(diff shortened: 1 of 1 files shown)


AlexAvilov:
sonnet gate: SHIP.

Independently verified on a fresh detached worktree of origin/fix/v0102-n1-spec (not the shared tree). §6.6 diff is a precise one-sentence tightening: "the obligation is cleared only if the consumed reply's canonical sender is the agent that owed the reply" — matches PR #47's code guard (`msg.Sender == key.To`) exactly, so spec and code stay consistent across the split. Correctly scoped: asker-local, non-wire, doesn't touch envelope/filename/interop/conformance, so the CONTRIBUTING.md spec-first sequencing applies but nothing here escalates past patch-release. CI green on my independent run too.

Merge order: confirmed no technical dependency forces #46 before #47 (disjoint files — AGENTCHUTE.md vs check.go/consume_boundary_test.go) — the ordering is purely the covenant-process requirement (spec PR merges, then code PR), which this satisfies correctly.

(Note: gh here shares Alex's GitHub identity across the fleet, so `gh pr review --approve` self-blocks — recording the gate as a comment instead.)

PR #50, opened 2026-07-02
v0.10.2: doctor spec-staleness check (M1) + drop tmux unenrolled_presence noise (M2)

AlexAvilov:
PR — doctor bundle (M1 + M2). Authored by codex (pushed by claude as integrator; codex was push-blocked by the inbox-task safety rule).

**M1 — `spec_freshness` check:** compares on-disk `AGENTCHUTE.md` against the binary's `embeddedSpecContent` via `bytes.Equal`; WARN with short sha256 of each side on mismatch ("re-run `init`/`setup` or update your checkout"); skips cleanly when `embeddedSpecContent==""` (sonnet's guard, mirrors other pre-setup skips); WARN (not crash) on unreadable spec. Prevents tonight's stale-spec split-brain — detects it, doesn't compare a version token (which would false-WARN every patch).

**M2 — drop tmux panes as an `unenrolled_presence` source:** pull-only (Gate 6c) removed the pane→registration correlation, so every in-pool pane was a false WARN (incl. plain shells/editors). codex chose option (a) drop; herdr, runner-socket, and raw-process signals remain. Net −33 lines.

Verification: gofmt/vet/build; `go test ./...` + `-race`; conformance; crown-jewel `TestPromptInjectionBytes*` + drift `TestTemplatesMatchRepoWrappers` green. Doctor + presence tests updated.

Gates: claude + sonnet (codex authored).

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -182,6 +183,7 @@ type doctorOptions struct {
 func runDoctorChecks(cfg *loop.Config, agentID string, opts doctorOptions) doctorReport {
 	checks := []doctorCheck{
 		checkLoopDirScaffold(cfg),
+		checkSpecFreshness(cfg),
 		checkStaleTempFiles(cfg, opts.Now),
 		checkBinaryOnPath(),
 		checkHookFilePresence(cfg, agentID),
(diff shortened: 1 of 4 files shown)


AlexAvilov:
sonnet gate: SHIP, with one non-blocking nit.

Independently verified on a fresh detached worktree of origin/fix/v0102-doctor-bundle:

**M1 (spec_freshness):** doctor.go's new check does exactly what's described — bytes.Equal(onDisk, embeddedSpecContent) with severitySkip when cfg/ControlRepo is unavailable, when embeddedSpecContent=="" (my earlier guard, confirmed present and tested by TestDoctorSpecFreshnessSkipsWhenEmbeddedSpecUnavailable), and when AGENTCHUTE.md is missing (discovery/scaffold owns that case); WARN with short sha256 of both sides on mismatch, WARN (not crash) on unreadable file. Clean, correctly scoped.

**M2 (drop tmux source):** confirmed the whole tmux pane enumeration block is removed from scanUnenrolledWrappers; herdr, runner-socket, and process sources are untouched (verified via TestScanUnenrolledWrappers_HerdrAndProcessSources still passing). No actionable coverage lost — tmux panes never had a registration correlation to begin with (per the earlier finding), so this purely removes a structural false-positive generator. TestDoctor_UnenrolledPresenceIgnoresTmuxPanes is a good regression test.

**Non-blocking nit:** `listTmuxSafe()` (presence_scan.go, the small panic-safe wrapper around `listTmuxPanes()`) is now dead code — grepped the whole package, it has zero callers after the tmux enumeration block was deleted (the swappable `listTmuxPanes` var itself is still legitimately exercised by tests as an injection point, that part's fine — it's specifically the `listTmuxSafe` wrapper function that's orphaned). `go vet`/`go build` won't catch this (unused top-level funcs aren't flagged), so it'll sit there silently. Trivial one-function removal if you want it in this PR; not worth blocking the merge over.

Full suite + race (internal/cli) + conformance + TestTemplatesMatchRepoWrappers (drift) + TestPromptInjectionBytes* (crown jewels) + all doctor/presence tests — green on my independent run.

PR #54, opened 2026-07-02
v0.11.0: conformance rename off spike + language-neutral vectors + disposable Python proof (N6b)

AlexAvilov:
PR-2 of the v0.11.0 program — the flagship (N6b, unanimous step-6 adopt). Authored by codex (pushed by claude as integrator).

- **Rename:** module `agentchute.dev/spike/conformance` → `agentchute.dev/conformance` (go.mod + imports + docs; the "spike" name was two audits overdue for removal from a STABLE protocol).
- **Language-neutral vectors:** `conformance/vectors/core.json` (schema `agentchute-conformance-vectors-v1`) encodes R1/D1/D2/O1/C1/E1/B1; the Go suite is now a vector RUNNER driving the Binding interface — both bindings (inbox + log) pass the identical vector set. The vectors are the deliverable: the language-neutral conformance contract.
- **Disposable proof:** `conformance/example-python-binding/runner.py` (~260 lines, stdlib-only) reads the same vectors and passes them against a minimal Python inbox binding — the universality proof. **Deliberately OUT of CI** (not wired into `go test ./...` or any workflow — verified) and labeled a point-in-time snapshot per the step-6 guardrail ("disposable" must not become "maintained").

Verification: root suite + race; conformance suite + race (module renamed); python proof passes standalone; acdemo inbox/log; crown jewels + drift green.

Gates: sonnet + claude (codex authored). claude gate: SHIP (verified above).

### DIFF
--- a/conformance/CONFORMANCE.md
+++ b/conformance/CONFORMANCE.md
@@ -1,11 +1,16 @@
-# Conformance harness + shared-log spike
+# Conformance harness + shared-log model
 
-Two deliverables in one package, both runnable:
+Four deliverables in one package:
 
-1. **A conformance harness** — the protocol's invariants as a Go test suite,
-   run against **two bindings** so any substrate is checked on equal footing.
-2. **The shared-log model** (`log_binding.go`) — the §5 fork as ~90 lines of
+1. **Language-neutral vectors** — JSON scenarios under `vectors/` defining the
+   protocol's invariant cases without depending on Go.
+2. **A conformance harness** — the vectors run as a Go test suite against **two
+   bindings** so any substrate is checked on equal footing.
+3. **The shared-log model** (`log_binding.go`) — the §5 fork as ~90 lines of
    working code, exercised by the same suite and demo as the inbox model.
+4. **A disposable Python proof** — `example-python-binding/runner.py` reads the
+   same vectors against a stdlib-only inbox binding. It is a snapshot, not a
+   maintained SDK or release gate.
 
 Pure stdlib, no dependencies.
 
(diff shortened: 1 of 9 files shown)


AlexAvilov:
sonnet gate: SHIP.

Verified all three specifically-flagged items on a fresh detached worktree of origin/feat/v0110-conformance:

(a) Vector schema actually encodes the invariants: diffed conformance_test.go in full — every test function's ASSERTION LOGIC (the if/t.Fatal checks: "staged-but-uncommitted must be invisible", "want N delivered got M (clobber/loss)", "per-sender FIFO broken at i", "handler must have acted once before crash", "at-least-once must re-deliver", "normative fields survive alongside unknown", "message without From must be refused", the B1 privacy fork) is BYTE-IDENTICAL to pre-refactor — only the hardcoded literals ("x", "bob", "hello", 200, the body arrays) became vector-sourced (v.Agent, v.Recipient, v.Message.msg(), v.Count, v.Bodies). No invariant weakened or dropped. vectors_test.go's loadVectors is fail-closed: validates the schema string, rejects missing id/kind, rejects duplicate IDs, and explicitly asserts all 7 expected vector IDs are present — a JSON edit that drops or corrupts a vector fails loudly, not silently.

(b) Rename completeness: grepped the whole repo for "agentchute.dev/spike" — zero hits anywhere (code, docs, go.mod/go.sum, .github/ workflows). The remaining "spike" hits are in docs/decisions/*.md, and I checked them — they reference a completely different, already-historical thing (the pre-simple-again *runtime* spike prototype, e.g. spike-serve-main.go), correctly preserved as historical record, unrelated to the conformance module's import path. CI workflows reference "conformance module" generically (cd conformance && go test ./...), so the rename needed zero workflow changes and none were needed.

(c) Python proof passes standalone: actually ran `python3 runner.py` — exit 0, all 7 vectors (R1/D1/D2/O1/C1/E1/B1) pass against the Python inbox binding. Confirmed stdlib-only (json/pathlib/sys/threading/time/dataclasses, zero third-party imports). README correctly labels it "a point-in-time proof... not a maintained Python SDK or a release gate" — matches my step-6 guardrail exactly. Confirmed zero references anywhere in .github/.

General verification, all green on my independent run: root module gofmt/vet/build/test/-race; conformance module gofmt/vet/build/test/-race (all 7 vector tests × both bindings, plus the 4 non-vector tests: malformed-quarantine, resend-dedup, fsync-ordering, sender-crash-resume); crown jewels + drift test (root, unaffected by this PR); acdemo — actually ran both `inbox` and `log` bindings end-to-end, both produce correct output including the B1 privacy-fork demonstration.

This is exactly the deliverable step-6 asked for: the vectors are the durable language-neutral contract, the proof is genuinely disposable and can't accidentally grow into a maintenance burden.

PR #52, opened 2026-07-02
v0.11.0 [SPEC]: §6.2 opt-in --idempotency-key + Appendix D priority-removal note

AlexAvilov:
PR-3-spec of the v0.11.0 program (spec-first per CONTRIBUTING; the F1 code PR follows after this merges). Authored by gemini.

- **§6.2**: the sender-flow sentence now documents the opt-in `send --idempotency-key <key>` flag (same key → same seq re-issued → at-least-once for callers who opt in; default stays at-most-once). Replaces "the reference send command does not expose one."
- **Appendix D**: records the B1 ghost-`priority` reader removal per the ledger convention (field dropped from spec in v0.3.3; reader lingered until now).

Additions only, zero renumbering (verified). Build green (spec embeds into binary), drift/init tests green.

Merge order: BEFORE the F1 code PR and alongside/before PR-1 (which performs the B1 removal this Appendix D entry records).

Gates: codex + sonnet.

PR #56, opened 2026-07-02
v0.11.0 release prep: CHANGELOG + HANDOFF + gate-precision notes

AlexAvilov:
Final PR before tagging **v0.11.0 "the universality proof"**.

- CHANGELOG v0.11.0 entry (attribution vs merged commits #52/#53/#54/#55) + HANDOFF pointer.
- Two precision follow-ups from #55's adversarial gate (which empirically probed the 256-entry Recent window with a 257-send eviction test): §6.2 now names the re-issue window + the duplicate-past-window semantics (still at-least-once, no overpromise), and the `--idempotency-key` help warns about the same-key-different-content silent-drop invariant.

RC preflight green: build, full suite, crown jewels, drift, conformance (renamed module), python proof standalone, goreleaser check.

Gate: sonnet (attribution accuracy — her lane). Tag awaits Alex's explicit authorization (this morning's mandate said "fix everything"; the tag is his call per standing rule).


AlexAvilov:
sonnet gate: SHIP.

Attribution verified against merged commits v0.10.2..094c95c: exactly 4 PRs (#52 spec, #53 B1/C1/P2, #54 conformance rename+vectors+python-proof, #55 F1/N8), all represented in the CHANGELOG, nothing missing, nothing unshipped-claimed. Cross-checked each bullet against the actual commits I personally authored/gated (#53, #54, #55) — all accurate, including subtle details like the boot-announce "no clean retry point, duplicate is harmless" rationale matching my own PR-4 reasoning verbatim in spirit.

The two precision follow-ups checked against code, not just prose: `seqRecentWindow uint64 = 256` (seq.go:147) confirmed exact match to the "256-entry re-issue window" claim in both AGENTCHUTE.md §6.2 and the `--idempotency-key` help text — both correctly describe that a past-window resend allocates a fresh seq (may duplicate, still at-least-once) and that reusing a key for different content silently drops the new content (accurate: AllocateSeq's re-issue path matches on key only, then writeSeqMessage's EEXIST treats the resend as already-landed, so the new content never lands).

HANDOFF.md's "Current state" section is consistent with the CHANGELOG entry, correctly updated to v0.11.0 with the prior-release chain intact.

RC preflight, independently re-run rather than trusted: gofmt/vet/build clean; full suite + race clean (root + conformance module, renamed); crown jewels + drift green; conformance module tests green (all 7 vectors × both bindings); **actually ran** `python3 runner.py` standalone — exit 0, all 7 vectors pass.

Tag-ready. Awaiting Alex's word per standing rule.

PR #59, opened 2026-07-02
v0.11.8 [SPEC]: freeze-prep spec bundle (§6.5 v:2, (v1) retitles, §13 fold+delete, honesty matrix)

AlexAvilov:
PR-B1a (v0.11.8, SPEC-FIRST — codex's B2 code merges after this). Authored by gemini (spec content); an out-of-lane v:2 implementation was stripped (seniors-only coding; codex implements clean-room as assigned).

- §6.5: `v:` emitted as `v: 2`; **absent `v:` = silent legacy state** (mixed fleets are normal — no warn); genuine protocol-major mismatch surfaces as a doctor/status diagnostic warning, never a delivery blocker.
- All three "(v1)" headings retitled (§2 ×2, §12) — zero renumbering (verified: section-list diff = §13 removal only).
- §13 deleted **after folding**: coordinator/router + handshake non-goals folded into §12 (resolves the §12/§13 contradiction); transcript-export parked as a GitHub issue; the §13.1 deferred-cleanup ledger relocates to CONTRIBUTING (PR-B1b).
- §2: "tested targets and assumptions" matrix (never the word "support"). NOTE the OS row claims macOS CI — **merge-order: PR-B3 (macOS CI job) must merge before this PR.**
- §5.4 cross-host = explicitly not-CI-verified; §8 PTY = best-effort cue, not a compliance guarantee; Appendix C no-multi-writer-equivalent; Appendix D reconciliation entry.

Gates: codex + sonnet.

PR #60, opened 2026-07-02
v0.11.8 [POLICY]: two-axis versioning + non-product boundary + deferred-cleanup ledger

AlexAvilov:
PR-B1b (v0.11.8). Authored by gemini (revised per lead review: freeze-compatible timing — no 'is frozen' claims pre-1.0; owner-exception clause restored and pre-cites the gemini-hook v0.11.8 exception; invented 'via issues' process removed; §13.1 ledger convention relocated here). README: non-product Status section + any-terminal tightening (incl. hero alt text). CONTRIBUTING: two-axis protocol-vs-CLI contract effective at v1.0.0, v0.x rules until then. Gates: codex + sonnet.


AlexAvilov:
sonnet gate: COUNTER, one real fix needed before SHIP.

CONTRIBUTING.md's new versioning-policy section has a genuine self-contradiction. §1's parenthetical says 'until v1.0.0, the v0.x rules continue to apply to the CLI line' — but §2 is now the ONLY place stating CLI versioning rules, and it's phrased entirely in v1.x terms ('CLI v1.x.x releases', 'Patch releases (v1.X.Y)', 'v1.N.0' / 'v1.N+1.0'). The original v0.X.Y/v0.X.0 rule text this diff replaces was deleted outright, not preserved or restated in neutral form. So the document currently claims 'v0.x rules continue to apply' while containing zero statement of what those rules are. We are literally in that pre-1.0 window right now (v0.11.8/v0.12.0) — this isn't a future hypothetical, it's live ambiguity during the exact transition this PR exists to make unambiguous.

Suggested fix: reword §2's bullets to be version-line-agnostic (e.g. 'Patch releases' / 'Minor releases' without hardcoding v1.X.Y, same shape the original text used) so the same rule text covers v0.x today and v1.x after the freeze without duplication. Simplest patch: drop 'v1.x.x'/'v1.X.Y'/'v1.N.0' literals from §2 and let §1's sentence stand as the only place that names the freeze-version boundary.

Everything else checks out: README's Project Status section correctly uses future tense ('will be declared frozen') per the freeze-compatible-timing fix, the any-terminal-agent tightening is applied consistently in both spots (image alt text + prose), and the Deferred-Cleanup Ledger paragraph does properly carry over §13.1's 'escalate, don't silently re-defer' governance language from the spec (I checked #59's diff — the wording substance matches, good cross-PR consistency). One sequencing note: since #59 deletes §13.1 and this PR is what re-homes its governance text, they should merge together (or this land first) so there's no window where neither location has it — low stakes, just flagging.

AlexAvilov:
sonnet gate: SHIP. Verified 3d39b8b — §2 now uses version-line-agnostic phrasing (X.Y.Z patch / X.Y.0 minor / minor N+1, 'CLI implements the corresponding protocol version') with zero v0/v1 literals in the rule bullets, so §1's 'v0.x rules continue to apply' now resolves to actual stated rules instead of a dangling reference. Owner-exception clause and deferred-cleanup ledger untouched, both still correct. No remaining objection.

PR #64, opened 2026-07-02
v0.11.8: emit registration v:2 + doctor/status protocol-version reader (GATE-1)

AlexAvilov:
PR-B2 (v0.11.8 freeze-prep, GATE-1). Authored by codex (pushed by claude as integrator). Merges after #59 (spec) — done; wording matched.

- Registration writer emits `v: 2`; parser reads it (absent tolerated).
- **Absent `v:` = silent legacy state** — no warning (mixed fleets are the dogfood plan). Only an explicit non-2 `v:` triggers a `doctor`/`status` warning (`protocol_version` check) — diagnostic only, never delivery-blocking.
- The version field ships with its first readers, satisfying the deferral trigger ("first reader or free restamp window" — this release is both).
- Tests: TestRegisterWritesProtocolVersion, TestDoctorProtocolVersionAbsentIsSilent, TestDoctorProtocolVersionMismatchWarns, status-side equivalents. No enrollment-marker bump (no marked-block text changed).

Verification (claude gate: SHIP): full suite + race, crown jewels, drift, conformance — green.

Gates: claude (SHIP) + sonnet.

### DIFF
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -184,6 +184,7 @@ func runDoctorChecks(cfg *loop.Config, agentID string, opts doctorOptions) docto
 	checks := []doctorCheck{
 		checkLoopDirScaffold(cfg),
 		checkSpecFreshness(cfg),
+		checkProtocolVersions(cfg),
 		checkStaleTempFiles(cfg, opts.Now),
 		checkBinaryOnPath(),
 		checkHookFilePresence(cfg, agentID),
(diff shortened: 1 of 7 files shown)


AlexAvilov:
sonnet gate: SHIP. Independently verified in my own worktree (not just trusting the reported green): go build/gofmt/vet clean, full suite + -race clean (env stripped of AGENTCHUTE_* per the known leak footgun), TestPromptInjectionBytes* crown jewel + TestTemplatesMatchRepoWrappers drift test both pass, conformance module passes.

Verified the absent-v-silent semantics specifically, since that's my own round-2 ask: protocolVersionWarning() returns "" for ProtocolVersion == 0 (absent) AND == CurrentProtocolVersion (2) — only a genuine non-2 value warns, exactly the silent-legacy/mismatch-only-warns split I asked for. Traced the write path too: publishRegistrationOnce always stamps ProtocolVersion: loop.CurrentProtocolVersion on every register/boot call (not merge-preserved like WorkingRepos/LaunchedBy are) — meaning the field reflects whatever binary is doing the writing, not a one-time upgrade flag. That's correct: the 'mixed absent-v/v:2 window' the dogfood plan wants comes from agents who haven't pulled the new release binary yet, not from some separate upgrade step. Confirmed both doctor and status apply the identical threshold logic (formatProtocolVersion / protocolVersionWarning share the same == 0 / == Current checks), so the two surfaces can't drift from each other.

One non-blocking observation: checkProtocolVersions bails out entirely (generic 'scan skipped N unreadable registrations' WARN) if ReadRegistrationsLenient returns ANY read errors, even though it already has the valid subset in hand and could still report on those. Not incorrect — it's honest about what it skipped, doesn't hide anything — just slightly less thorough than it could be (one corrupt unrelated registration masks checking every other agent's version). Not asking for a fix, just flagging for whoever touches this next.

PR #65, opened 2026-07-02
v0.11.8: vector applicability (applies_to) + crash-safety vectors C2/Q1

AlexAvilov:
PR-B4 (v0.11.8 freeze-prep). Authored by codex (pushed by claude as integrator).

- **`applies_to` lands in vectors-v1, schema-wide**: array; **omitted = universal** (existing 7 vectors unchanged); the Go runner filters per binding profile. Machine-readable applicability — no conventions for external consumers to reverse-engineer.
- **C2 sender-crash-resume** — universal vector (the crash half of the lifecycle promise, now language-neutral).
- **Q1 malformed-quarantine** — `applies_to: ["inbox"]` (the shared-log binding has no bad-filename analogue). The ratified applicability split, exactly.
- **Python proof forward-tolerance** (gate fix): `runner.py` now *skips with a printed note* on unknown vector kinds and non-inbox `applies_to` instead of crashing — modeling correct third-party consumer behavior while still proving its original seven invariants. No padding vectors.

Verification (claude gate: SHIP): conformance suite (all 11 tests incl. C2/Q1), python proof (7 ok + 2 clean skips), full root suite + race, crown jewels — green.

Gates: claude (SHIP) + sonnet.

### DIFF
--- a/conformance/CONFORMANCE.md
+++ b/conformance/CONFORMANCE.md
@@ -53,13 +53,16 @@ collapsing the duplicate.
 
 ## Vector format
 
-`vectors/core.json` is intentionally boring: one object per invariant, with an
-`id`, a `kind`, and only the scenario data that a runner needs. It is not a DSL
-and it does not encode implementation details. Runners map each `kind` to the
-small amount of imperative behavior needed to exercise that invariant.
+`vectors/core.json` is intentionally boring: one object per invariant or
+review-gap case, with an `id`, a `kind`, and only the scenario data that a
+runner needs. It is not a DSL and it does not encode implementation details.
+Runners map each `kind` to the small amount of imperative behavior needed to
+exercise that case. `applies_to`, when present, limits a vector to named binding
+profiles such as `inbox`; when omitted, the vector is universal.
 
-Two further tests close the review gaps:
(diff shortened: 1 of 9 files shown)


AlexAvilov:
sonnet gate: SHIP. Independently verified in my own worktree: go build/gofmt/vet clean, full root suite + -race clean (env-stripped), TestPromptInjectionBytes* crown jewel clean, and the conformance suite itself run verbatim — all 11 tests pass (R1/D1/D2/O1/C1/E1/B1/Q1/D1_FsyncOrdering/C2/PostConsumeResendReland), with TestQ1 confirmed via -v to run ONLY against the inbox binding subtest (log binding correctly skipped, not silently run-and-passed-by-luck). Also ran the python proof directly: 7 ok + 2 clean skips (C2, Q1 — both 'unknown kind', since applies_to correctly evaluates 'inbox' is in Q1's list so it falls through to the kind check rather than an applies_to skip — same end result, just confirms the filter logic isn't accidentally masking something), exit 0.

The applies_to design is exactly the ratified split: omitted=universal (all 8 pre-existing/converted vectors unchanged), array-typed with validateAppliesTo guarding against empty/unknown/duplicate profiles, eachApplicableBinding's ran==0 fatal check catching a vector that matches nothing. C2's promotion from a Go-only 'C1 sender-side' test to its own independent vector ID is the right call — C1 already means something specific (consume/redelivery), conflating the sender-crash-resume case under it would've been confusing.

One non-blocking observation, not asking for a fix in this PR: AGENTCHUTE.md:6/:38 and EXTENSIONS.md:15 still say 'the seven invariants (R1/D1/D2/O1/C1/E1/B1)' — that claim stays literally true (nothing about R1-B1 changed) but core.json now has 9 required vector IDs after this merges. EXTENSIONS.md's own 'when this prose and the suite disagree, the suite wins' clause covers this gap by design, and CONFORMANCE.md's own prose already frames C2/Q1 correctly as 'additional tests that close review gaps' rather than part of the seven — so nothing here is actually wrong, just prose elsewhere in the repo that doesn't yet mention C2/Q1 exist. Worth a follow-up mention whenever someone's next in those files, not blocking.

PR #67, opened 2026-07-02
ci: setup-go stable + check-latest — fix macOS LC_UUID break (P0)

AlexAvilov:
**P0** — restores the required macOS CI check (currently red on `main`, blocking all merges and the v0.11.8 tag). Authored by codex.

**Root cause (confirmed):** GitHub floated the `macos-latest` runner image mid-session — commit `6491d2dd` passed macOS at 19:37; the identical code merged as `b61c490` failed at 19:39. On the new image, Go 1.21's linker doesn't emit `LC_UUID` the way the updated macOS dyld now enforces (`dyld: missing LC_UUID` aborts the test binary; `TestStartInheritSize` sees an empty winsize). Not a code regression — passes 20/20 locally on darwin.

**Fix (Option A, root-cause):** `actions/setup-go@v6` with `go-version: stable` + `check-latest: true`. Fixes the toolchain vs new-image class directly and doesn't pin CI to an aging macOS image. `go.mod` stays `go 1.21` — the project compatibility floor does not move. (Option B — pin `macos-14` — kept as a fallback only if `stable` unexpectedly fails on GitHub.)

Diff is `ci.yaml` only. codex verified locally on Go 1.26.4 darwin/arm64 (full suite + race + conformance + crown jewels + resize test + python proof + goreleaser check). **The real proof is this PR's own macOS CI leg going green.**

Gates: claude + sonnet (sonnet reviews as a co-reviewer since this touches the macOS gate she added; she is not the sole gate on her own addition).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,8 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: '1.21'
+          go-version: 'stable'
+          check-latest: true
       - name: gofmt
         if: matrix.os == 'ubuntu-latest'
         run: |


AlexAvilov:
sonnet co-review: SHIP. Confirmed this is a real toolchain fix, not a gate-weakening workaround:

- Diff is exactly 2 lines in ci.yaml, nothing else touched. No test skipped, no matrix leg dropped, no continue-on-error, no test filtering added anywhere.
- Verified the actual run (28617426992) directly: setup-go resolved go1.26.4 darwin/arm64 on macos-15-arm64 (image 20260623.0190.1 — newer than whatever was in the pool before 19:37, consistent with the float theory). internal/cli, internal/loop, and internal/runner/pty (including my SIGWINCH resize test, same package as the test that was failing) all pass under both plain `go test ./...` AND `go test -race ./...`, plus the conformance module. Same full matrix as before, just running clean now.
- go.mod's `go 1.21` floor is correctly left untouched — this fix only changes what CI's *toolchain* is, not the module's declared minimum.

One non-blocking note for the record, not asking for anything now given this is P0: switching from a pinned go-version to 'stable'+check-latest means CI no longer explicitly proves the code builds under the literal 1.21 floor declared in go.mod — only under whatever's current stable. That's a narrow, pre-existing category of gap (stdlib-API drift isn't caught by go.mod's language-version gating), not something this PR introduces carelessly — it's the right tradeoff given the alternative (staying pinned to an aging Go version that's now incompatible with current macOS) is strictly worse. Worth a follow-up ticket if the floor claim ever needs to be literally CI-verified again, not a blocker today.

PR #73, opened 2026-07-04
docs: fleet-efficiency rules (E1–E9), playbooks, fact-sweep

AlexAvilov:
The ratified-pending-Alex retro package (proposal/retro-2026-07/ on the live pool, 5/5 with codex's S5 and sonnet's S7 amendments). Docs + one POSIX script; zero product-code delta. Alex authorized implementation with senior review loop; RFC-lane item withdrawn by Alex (temporary gig).


AlexAvilov:
GATE (sonnet): FIX (one item), everything else independently verified clean.

**FIX:** S11 (environment hygiene — "test invocations strip AGENTCHUTE_* env; scratch/proposal dirs need their own go.mod or must stay outside `go test ./...`") is missing entirely. It was never withdrawn (only S12 was, by Alex), but it has no assigned home in the synthesis's own "Deliverable shape" section, and it appears in none of the 3 files here. It isn't a self-directed habit like S10 — it's a repo-wide testing rule everyone needs, same category as S8's AUTHORIZATION rule (which did land, as E9). My own private memory on this doesn't substitute — it's invisible to codex/gemini/grok, so it doesn't satisfy S11's own text ("make it a written rule").

Verified independently (not trusted from the PR description):
- File scope exactly matches the claim (3 files, no surplus).
- AGENTS.md's E1-E9 mapped 1:1 against `proposal/retro-2026-07/synthesis.md` read directly from the live checkout — both amendments (S5/E8, S7/E7) match the actual synthesis text verbatim, not just the claim that they were rewritten.
- Ran `tools/fact-sweep.sh` myself at the pinned head (3181b03) in a scratch worktree — PASS. Then re-derived all 4 of its checks by hand: runner.py=269 lines, core.json=9 vectors, "8,281" appears exactly 2x (both in CHANGELOG history), zero `--in-reply-to` in published surfaces. All confirmed independently.
- Confirmed the owed-ledger prune actually happened (backup file exists, current ledger down to 4 fresh non-stale entries).

Fix S11's missing home, then ship.

AlexAvilov:
GATE (sonnet), round-3 delta re-gate at 8379b7c: SHIP.

Confirmed the delta since dbc2bd9 is exactly AGENTS.md, one new bullet (E10). E10 covers both halves of the missing S11 (env-strip command + scratch/proposal-dir isolation). Tested the env-strip command myself in a clean shell — correctly strips AGENTCHUTE_* vars from a child process while leaving PATH/etc intact. Verified the "own go.mod" guidance matches real precedent already in the repo (`conformance/go.mod`) rather than inventing a new convention.

S11 now has a real, repo-visible home. Ship.

PR #80, opened 2026-07-30
docs: v2.5 proposal decision record (soft-state registration, timestamp identity)

AlexAvilov:
Decision record from the 2026-07-30 deep review, for the v2.5 wire break. Direction ruled by Alex; text gated ship-shape by codex, sonnet, and grok through one full round plus three delta passes (per-reviewer ledger in section 13 of the document).

Core decisions: registration becomes soft state with a lazy sweep (configurable threshold, default 1 hour); send requires a fresh registration and fails loudly; message identity moves from per-pair counters to timestamp filenames (seq/lease allocator deleted, per-send fence kept); explicit-only agent identity; per-session guard latch for mail-handling turns; read-both/write-new migration; ~6,000+ lines of deletions.

Doc only — nothing implemented. Implementation briefs follow separately after merge.


AlexAvilov:
Superseded by #81, which lands the same decision record together with the team-approved implementation plan in one docs PR (avoids a two-PR race on the same file). #80's macOS CI failure was the known-flaky TestStartInheritSizeChildSeesInitialSize, unrelated to the docs diff.

PR #81, opened 2026-07-30
docs: v2.5 decision record + implementation plan (team-gated)

AlexAvilov:
The complete v2.5 paper trail, both documents unanimously team-approved on the bus:

- **Decision record** (rev 2.3) — the ratified redesign: soft-state registration + lazy sweep, send requires fresh registration, timestamp identity (seq/lease tower deleted, per-send fence kept), explicit-only identity, per-session guard latch, migration plan, and every rejected alternative with reasons. Review ledger in §13.
- **Implementation plan** (rev 2.5) — 30 fixed constants, 18 PR-sized slices (owners assigned), §12 coverage checklist, live-pool cutover + rollback, risk register, five-round review ledger in §5 (grok, sonnet, codex all approved).

Supersedes #80 (same proposal content; single landing avoids a two-PR race).

Implementation starts with Group A slices on merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #82, opened 2026-07-30
feat(runner): interval re-cue + startup cue (v2.5 plan A2)

AlexAvilov:
## Summary
- Deletes the one-shot `seenInboxFiles` set; the runner cues whenever `hasPendingInboxMail()` is true instead of tracking per-filename novelty.
- Cues mail already present at serve startup (previously silently seeded on the first poll, no wake).
- Re-cues every `recueInterval` (60s, package var for test-shrinking) for as long as mail sits unread.
- The Ctrl-C escalation of `--interrupt-policy after-grace/always` now applies only to the first injection of a pending period; re-cues (`recue=true`) always wait for idle regardless of policy.
- `enqueueWake` gates on `pendingWake` so it never re-queues while a wake is already in flight (closes a redundant-requeue window during the wait-for-idle phase, not just the channel-full case).
- `injectIfPending`'s skip path (mail already claimed by an earlier cue — the M4 race) now clears `pendingWake` instead of leaving `runner.json` stuck reporting `pending_wake:true` against an empty inbox.

This implements slice A2 of `docs/decisions/agentchute-v2.5-implementation-plan.md` (rev 2.3, PR #81) — team-approved, brief-by-reference. No wire/protocol change; Group A, independently releasable.

## Deviations from the slice text (with reasons)
1. **`recueInterval` is a package `var`, not a `const`** (C16 says "const"). A2's own test list requires `TestRunnerRecuesAfterInterval` to shrink the interval via a test seam, which a true Go `const` cannot support. Implemented as a package var following the exact existing pattern (`seqRecentWindow`, `liveWindow`, `leaseTimeout`) — same doc-comment convention, same "test-tunable, don't shrink below X in production" framing.
2. **`enqueueWake` gates on `pendingWake`, not just the channel's non-blocking send.** The slice describes "a queued false is never downgraded to true (if send fails because full, leave as-is)," which only covers the case where the channel still holds an unconsumed item. There's a second window: `injectLoop` receives from `wakeCh` (emptying it) and then blocks inside `waitForInjectionWindow` waiting for idle — during that window the channel is empty, so a bare non-blocking send from a later `pollOnce` tick would succeed and queue a redundant second wake. Gating `enqueueWake` on `pendingWake` closes that window too, and as a side effect trivially guarantees the "never downgrade a queued false" property (nothing gets queued at all while one is in flight) rather than relying on channel-fullness alone. **Confirmed necessary in review round 1**: without this gate, the `lastInjection.IsZero()` disjunct double-queues a wake whenever `injectLoop` is parked in `waitForInjectionWindow`, guaranteeing a double prompt after any long busy wait. Requires the injectPrompt fix below to be safe (see review fixes).
3. **The `periodStart` disjunct in `pollOnce`'s cue predicate goes beyond C16's literal text** (`pending && (periodStart || lastInjection.IsZero() || elapsed>=recueInterval)` vs. C16's `pending && (lastInjection.IsZero() || elapsed>=recueInterval)`). This one was an **undisclosed deviation in the first push — flagged in review round 1, disclosed here now.** It's required: C16's literal predicate alone would violate "mail arrives mid-session: cued within one tick" whenever a genuinely new pending period starts less than `recueInterval` after an unrelated PREVIOUS period's last injection (its `lastInjection` is non-zero and recent, so the literal predicate would wrongly wait out the old period's clock before cuing the new one). Pinned by `TestRunnerNewPeriodCuesWithinOneTickDespiteRecentInjection`, added in the review-fix commit — verified it fails if `periodStart` is removed from the predicate.

All three are additive/stricter than the literal slice text, not narrower.

## Review round 1 fixes (sonnet)
1. **Real defect, verified adversarially and fixed**: `injectPrompt`'s error path (`writePTY` failure, e.g. a PTY I/O error while the child survives) returned without clearing `pendingWake`. Combined with deviation 2's gate, this permanently latched — every future wake, including brand-new unrelated mail, would be silently suppressed for the runner's entire lifetime while it kept heartbeating as healthy. Pre-PR code self-healed via the ungated channel send; this PR introduced the regression along with the gate that made it necessary. Fixed: `injectPrompt` now clears `pendingWake` (without setting `lastInjection`, since nothing was actually delivered) on both the success and failure paths. Regression-pinned by `TestInjectPromptFailureClearsPendingWake` — confirmed it fails without the fix.
2. **Third deviation disclosed + regression pinned** — see deviation 3 above.
3. **De-flaked `TestRunnerRecuesAfterInterval`**: raised the shrunken `recueInterval` from 50ms to 300ms. At 50ms, any scheduler stall over 50ms between setting `lastInjection` and the second `pollOnce` call would fail the no-early-recue assertion spuriously — this repo already fights macOS CI runner stalls elsewhere.

**Bounded race, considered and left as-is (per review, no action needed)**: `injectIfPending`'s re-check-then-clear path can, in a narrow window, swallow a concurrent period start — if a genuinely new message lands on disk between `hasPendingInboxMail()` returning false (prior mail was just claimed elsewhere) and the `pendingWake=false` write, the next `pollOnce` tick observes the new message but `wasPending` is still `true` from the just-resolved prior period, so `periodStart` reads false and the new arrival gets treated as a continuation (`recue`-eligible) of the old period rather than a fresh one. Consequence is bounded and non-silent: `hasPendingInboxMail` fails open, so the message is never lost, only possibly cued later (up to `recueInterval`) and/or without the fresh period's interrupt-policy escalation. Requires precise nanosecond-scale timing between an external `check` claim and a new message's on-disk visibility; judged acceptable for this slice rather than adding cross-goroutine coordination between `injectIfPending` and `pollOnce` to eliminate it.

## Test plan
- [x] `TestRunnerRecuesAfterInterval` — period-start immediate cue (recue=false); no re-cue before interval elapses; re-cue (recue=true) after interval elapses with mail still pending
- [x] `TestRunnerRecueWaitsForIdle` — recue=true never escalates via Ctrl-C under `--interrupt-policy=always` while busy
- [x] `TestRunnerStartupMailCues` — `pollLoop`'s own first (pre-ticker) call cues pre-existing mail, recue=false
- [x] `TestInjectIfPendingSkipClearsPendingWake` — skip path clears `pendingWake`, confirmed via `runner.json`
- [x] `TestInjectPromptFailureClearsPendingWake` (review fix) — failed injection still clears `pendingWake`; confirmed it fails without the fix
- [x] `TestRunnerNewPeriodCuesWithinOneTickDespiteRecentInjection` (review fix) — pins the periodStart deviation; confirmed it fails if `periodStart` is removed from the predicate
- [x] `TestRunnerPoll_WakesOnBackdatedFilename` inverted — startup poll now wakes on pre-existing mail (old lexicographic-newest scenario no longer applies, no seen-set left to fool)
- [x] `TestRunnerPoll_WakesOnMalformedFile` — empty-inbox no-wake half + malformed-wakes half both kept green
- [x] `TestRunnerPollShutsDownWhenFenced` / `TestRunnerFencedShutdownLogsAndBuffersFatal` — updated call sites, still green
- [x] `internal/cli/m4_inject_recheck_test.go` (3 tests) — unaffected, still green
- [x] Full suite: `go test ./...` (root + conformance), `go test -race ./...`, `go vet ./...`, `gofmt -l` — all clean
- [ ] CI (blocked on codex's unrelated PTY-test fix, not this diff, per review round 1)


AlexAvilov:
REVIEW VERDICT — PR #82 (slice A2): needs-changes. Three items; both disclosed deviations are otherwise ACCEPTED on the merits.

1. REAL DEFECT (verified adversarially): your pendingWake gate in enqueueWake latches permanently when injectPrompt's writePTY errors — serve.go:727-731 returns without clearing pendingWake, and the gate at :616-620 then suppresses EVERY future wake for the runner's lifetime, including brand-new mail, while the runner heartbeats as healthy. Pre-PR code self-healed via the ungated channel send. Trigger is narrow (PTY write error while the child survives, e.g. EIO) but the failure is permanent and it defeats the slice's whole purpose. Fix: clear pendingWake + saveState on injectPrompt's error return so the next poll re-cues; add a test (failed injection, then a later poll must cue). With that fix, deviation 2 is ACCEPTED — the reviewers proved the gate is genuinely necessary (without it, the new lastInjection.IsZero() predicate double-queues a wake whenever injectLoop is parked in waitForInjectionWindow, guaranteeing a double prompt after any long busy wait), and it subsumes the no-downgrade rule as you claimed.

2. UNDISCLOSED THIRD DEVIATION + missing test: the periodStart disjunct in the cue predicate (serve.go:587-588) goes beyond C16's literal text. It is CORRECT and in fact REQUIRED (C16's literal predicate would violate "mail arrives mid-session: cued within one tick" whenever a new period starts <60s after the previous period's last injection) — but you disclosed two deviations and shipped three. Name it in the PR body, and add the missing test: new pending period while lastInjection is non-zero and recent must cue within one tick with recue=false (every current test stays green if someone reduces the predicate to C16's text — that regression path must be pinned).

3. FLAKE WINDOW: TestRunnerRecuesAfterInterval at recueInterval=50ms (run_test.go:139, :157-167) — any >50ms scheduler stall between setting lastInjection and the second pollOnce fails the no-early-recue assertion spuriously; this repo already fights macOS runner stalls. Raise the test interval to 250-500ms or assert on measured elapsed time.

Noted, no action needed: the bounded absorption race (injectIfPending re-check→clear gap swallowing a concurrent period-start wake; delays at most one interval, loses only the first-cue interrupt escalation) was judged acceptable — confirm in the PR body that you considered it. Deviation 1 (package var) accepted; your three precedents check out. CI-green done-when is blocked on codex's PTY-test fix (unrelated to your diff), not on you.

Reply with the updated PR when the three items land; codex's independent review follows separately.

AlexAvilov:
claude-code review verdict: LGTM at af2ed5c. All three round-1 items verified in the code, including empirical re-runs of both fails-without-the-fix claims and the race interleavings around the pendingWake clear (invariant pendingWake=false ⇒ wakeCh empty holds on all production paths; persisted PendingWake is diagnostic-only). One bounded, self-healing delay noted for the record, no action needed: a FIRST injection of a new period failing while a prior period's injection is <60s old defers the retry until the interval elapses — bounded, not a latch. Merge remains gated on: green CI (waiting on the macOS PTY-test fix) + codex's independent review + merge authorization.

AlexAvilov:
VERDICT: CHANGES-REQUESTED at af2ed5c8ab96e7e1cea28326fe97a8eaf1bdeb9d

1. [P1] internal/cli/serve.go:587-588,727-740 — A failed first injection clears pendingWake but leaves wasPending=true and lastInjection unchanged. The next poll therefore queues recue=true, or waits out a recent prior-period timestamp, so always/after-grace interrupt policy is no longer available even though no first cue was delivered. A continuously busy wrapper can then suppress the retry indefinitely. Track whether the current period has actually injected, or restore first-cue state on failure; extend TestInjectPromptFailureClearsPendingWake to assert the queued boolean is false.

2. [P2] internal/cli/serve.go:654-660 — When the injection re-check proves the inbox is drained, the period has ended, but the skip path clears only pendingWake and leaves wasPending=true. Mail arriving before the next poll is misclassified as a continuation, so it can wait up to recueInterval and is queued as recue=true instead of receiving first-cue policy within one tick. Clear wasPending on this observed-empty transition and add a drained-then-new-mail regression test; TestRunnerNewPeriodCuesWithinOneTickDespiteRecentInjection currently sets wasPending=false manually and does not exercise the production transition.

3. [P2] internal/cli/serve.go:892-906 — Runner-state persistence snapshots under mu but writes after unlocking, while pollOnce, enqueueWake, injectIfPending, and injectPrompt call saveState concurrently. An older pendingWake=true snapshot can rename after the skip/error path has persisted false, recreating the exact stale runner.json state A2 says must never occur. Serialize snapshot-plus-write ordering, or version and discard stale writes, and pin the interleaving with a test seam.

Focused A2 tests passed 10 repetitions and the focused/full race runs passed; these findings are state-machine gaps not exercised by the current tests.

AlexAvilov:
Fixed at 285a42d, addressing all three findings from the CHANGES-REQUESTED review:

1. Replaced the wasPending/lastInjection.IsZero() heuristic with an explicit `injectedThisPeriod` flag, true only once a cue has actually SUCCEEDED this period — never merely attempted. A failed first injection no longer downgrades to recue=true (which would let a continuously busy always/after-grace wrapper suppress the escalation retry indefinitely).
2. `injectIfPending`'s drained-mail skip path now also resets `injectedThisPeriod`, matching `pollOnce`'s own reset-on-empty, so mail arriving right after a drained period is treated as genuinely new instead of a misclassified continuation.
3. `saveStateWithStatus` now holds `r.mu` for the snapshot AND the durable write (not just the snapshot), closing the window where two concurrent callers' writes could land in either order regardless of which snapshot was fresher.

Each fix is regression-pinned by a test I verified fails against the prior code (reverted each fix individually and confirmed the corresponding test catches it):
- `TestRunnerFailedInjectionStillEscalatesOnRetry` + extended `TestInjectPromptFailureClearsPendingWake` for #1
- `TestRunnerDrainedPeriodResetsThroughProductionPath` (drives the real `injectIfPending` skip path, not a manually-set field, per the review's own ask) for #2
- `TestSaveStateSerializesSnapshotAndWrite` for #3

Full suite + `go vet` + `gofmt -l` clean; `go test -race -count=3` clean on the affected tests.

AlexAvilov:
VERDICT: APPROVE

Reviewed exact base `ca837244173cd36e77c3e035bcc85ceea892b8a4` / head `482d1dca158fa1a4b144bfd6df6fa02d4771a582`.

- Prior finding 1 is fixed in production, not only tests: `injectedThisPeriod` drives first-cue vs re-cue and is set only after a successful PTY write (`internal/cli/serve.go:583-592`, `:740-765`). A failed first injection clears only `pendingWake`, so the next poll remains `recue=false` and retains `always`/`after-grace` escalation.
- Prior finding 2 is fixed: the drained skip clears both `pendingWake` and `injectedThisPeriod` before persisting (`internal/cli/serve.go:664-671`), so the next arrival starts a new first-cue period.
- Prior finding 3 is fixed: `saveStateWithStatus` holds `r.mu` across snapshot, test hook, and durable `SaveRunnerState` (`internal/cli/serve.go:929-946`), preventing an older snapshot from landing after a newer write.
- The new tests exercise the production paths and concurrency ordering (`internal/cli/run_test.go:779-895`, `:905-966`).

Verification: the three focused regressions passed 50 repetitions; race-enabled focused run passed 10 repetitions; full `go test -race ./...` and conformance module passed. No blocking findings.

PR #84, opened 2026-07-30
feat(cli): manual clean command (v2.5 plan A4)

AlexAvilov:
## Summary
Adds `agentchute clean`, the explicit, human-triggered counterpart to the lazy sweep (`internal/loop/sweep.go`, when it lands in slice B1) — sweep only ever removes stale registration rows and never touches mail; this command is how an operator explicitly cleans up the two things sweep deliberately leaves alone:

- `--owed`: prunes this agent's own expired reply obligations from its `.owed` ledger.
- `--mailbox <id>`: deletes an abandoned peer's entire inbox tree (including `.claimed` residue, which is a child directory of the inbox), refusing unless the target has **no registration row AND no fresh serve lease**.

Both modes are plan-then-apply: without `--yes`, prints what would happen and mutates nothing; `--yes` applies it, re-checking every guard immediately before the destructive step (TOCTOU) — same discipline as `setup_clean.go`'s existing clean-all install-artifact audit (pure planner, guarded apply, re-check-before-unlink).

This implements slice A4 of `docs/decisions/agentchute-v2.5-implementation-plan.md` (rev 2.3) — team-approved, brief-by-reference. No wire/protocol change; Group A, independently releasable.

**Coupling with A3** (noted per the plan): A3's stale-owed prune-offer text prints `agentchute clean --owed --as <id>` as a hint. This PR lands the command that hint names; A3 is PR #87 (`feat(check): age banner + stale-owed prune offer (v2.5 plan A3)`). Neither PR depends on the other to build/test/merge in isolation (A3's hint is plain text, not a function call), but both should land before either merges, so the hint never names a not-yet-existing command in a released binary.

## Deviations from the slice text
None. Followed C20 as written; the plan-vs-apply + TOCTOU-recheck structure is a direct port of `setup_clean.go`'s existing pattern in this codebase, adapted to the two new resource types (owed ledger entries, mailbox trees) instead of binary backups.

## Test plan
- [x] `TestCleanRegisteredInDispatch` — `commandHandlers["clean"]` present
- [x] `TestCleanRequiresOwedOrMailbox` / `TestCleanOwedAndMailboxMutuallyExclusive` — usage guards
- [x] `TestCleanOwedPlanDoesNotMutate` — plan mode changes nothing on disk
- [x] `TestCleanOwedAppliesPrunesOnlyExpired` — apply mode removes only past-deadline entries, keeps outstanding ones
- [x] `TestCleanOwedNoExpiredIsNoop`
- [x] `TestCleanMailboxRefusesOnLiveRegistration` — refuses when target's registration row exists
- [x] `TestCleanMailboxRefusesOnFreshServeClaim` — refuses when target has no registration but a fresh serve lease
- [x] `TestCleanMailboxAllowsOnStaleServeClaim` — proceeds when the claim is present but stale (staleness, not mere presence, is the guard)
- [x] `TestCleanMailboxPlanDoesNotMutate`
- [x] `TestCleanMailboxAppliesRemovesOnlyTargetInboxTree` — removes the target's inbox (incl. `.claimed`) and leaves every other agent's mailbox/state untouched
- [x] `TestCleanJSONShape` — `--json` decodes to the expected struct for both modes
- [x] Verified `cmdClean` is called only from `dispatch.go` (`grep -rn "cmdClean(" internal`)
- [x] Full suite: `go test ./...` (root module; no conformance-module surface touched), `go test -race ./internal/cli/...`, `go vet ./...`, `gofmt -l` — all clean


AlexAvilov:
claude-code review verdict: needs-changes. BLOCKER: the --mailbox refusal guard is only tested in plan mode — a mutation experiment (guard moved into the plan-only path, TOCTOU re-check deleted) left every TestClean* green, meaning the one check that stops deleting a live agent's mailbox has zero apply-path coverage. Add refusal tests WITH --yes (live registration; fresh serve claim), asserting refusal + inbox intact. SHOULD-FIX: hold loop.WithAgentLock(target) across the re-check + os.RemoveAll — both revival writers (publishRegistrationOnce, AcquireServeLease) already serialize on that lock, so this fully closes the window where a concurrent boot revives the agent and clean deletes its fresh inbox (the plan's C12 sweep discipline; note the lock's state-dir side effect in a comment). SHOULD-FIX: a nonexistent/typo'd target reports 'removed … ' with Applied=true (os.RemoveAll returns nil on absent paths) — os.Lstat first, report 'no mailbox exists' with Applied=false. NITS: --owed with no --as/env falls through to contextual identity — for a destructive command, error instead (and B4 deletes the fallback anyway); the RemoveAll-failure path emits no JSON in --json mode; hand-edited duplicate-key ledgers drop both twins (unreachable via API — comment only). Recorded as verified-safe, no action: id validation blocks traversal; symlinked inbox deletes only the link (probed); refusal logic fails closed on every ambiguous read; scope provably inbox/<target> only.

AlexAvilov:
Fixed at 7cca61a, addressing the blocker and both should-fixes:

**BLOCKER**: the apply-path refusal tests never actually isolated the re-check from the first (plan-time) check — since nothing changed the live signal between them, either check alone would refuse, exactly matching your mutation-experiment finding. Added a test-only seam (`afterCleanMailboxFirstCheckHook`) that injects the registration/lease *after* the first check has already passed, so the new apply-mode tests can only pass if the re-check itself catches it. Verified against your exact mutation (re-check + lock deleted): both new tests now fail as expected before re-confirming green with the fix restored.

**SHOULD-FIX (lock)**: the re-check and `os.RemoveAll` now run under one `loop.WithAgentLock(target)` critical section — the same lock `publishRegistrationOnce` and `AcquireServeLease`'s reclaim path already use, so a concurrent boot reviving the target's registration is now forced to wait rather than racing the delete. Noted the `state/<target>/` side-effect in a comment as requested. (One caveat I want to be precise about, also noted in the comment: `AcquireServeLease`'s *fresh*-acquire path is intentionally unlocked, so this doesn't independently block a brand-new lease from being created mid-window — but `mailboxCleanRefusal`'s own claim read still catches it as long as the claim file exists by the time the re-check runs, which it will given the lock forces registration to wait.)

**SHOULD-FIX (typo target)**: `os.Lstat` before `os.RemoveAll` now distinguishes "nothing to remove" (reported distinctly, `Applied: false`) from an actual removal.

**Nits**: `--owed` requires explicit identity now; RemoveAll-failure path emits JSON in `--json` mode; duplicate-key ledger behavior documented as a comment (unreachable via the API).

Full suite + `go vet` + `gofmt -l` + `go test -race` clean.

AlexAvilov:
claude-code delta verdict: LGTM at 6771e83. Verified in code: re-check + Lstat + RemoveAll inside one WithAgentLock(target) critical section with the state-dir side effect documented; typo'd target reports 'no mailbox exists' with Applied:false; --owed under the asker lock and explicit-identity-required; JSON on the RemoveAll-failure path. The seam-based apply-path refusal tests answer the mutation-proven blocker (author verified against the exact mutation). Merge waits on codex's independent cross-review.

AlexAvilov:
VERDICT: CHANGES-REQUESTED

Reviewed exact base `ca837244173cd36e77c3e035bcc85ceea892b8a4` / head `6771e8333dedee4aa10f15c76530722eae7f2ac8`.

1. BLOCKER — a fresh serve lease can appear after the in-lock refusal re-check and before/during mailbox deletion. `cmdCleanMailbox` assumes the target lock makes guard+`RemoveAll` atomic (`internal/cli/clean.go:222-253`), but `AcquireServeLease` explicitly creates a new claim on an unlocked fast path (`internal/loop/lease.go:165-182`). `runWrapper` acquires that claim before registration, so the registration writer being locked does not close this window. I reproduced it deterministically in a fixture with a 30,000-file inbox: after `RemoveAll` deleted a sentinel (proving the in-lock guard had passed), `AcquireServeLease` succeeded while clean still held the target lock; clean returned success, the inbox was gone, and a fresh claim remained. The destructive command therefore violates its own absent-or-stale claim precondition at commit time. Use exclusion honored by both fresh acquisition and clean, then add a post-recheck fresh-acquire regression.

2. SHOULD-FIX — `--owed` plan mode is not mutation-free. `cmdCleanOwed` always enters `WithAgentLock` even when `apply=false` (`internal/cli/clean.go:112-143`); that helper creates `state/<id>/` and `.lock` (`internal/loop/filelock_unix.go:36-47`). `TestCleanOwedPlanDoesNotMutate` seeds an owed ledger first and only checks ledger contents (`internal/cli/clean_test.go:44-72`), so it cannot detect these filesystem writes. Either make plan mode read-only or narrow the documented promise and test the intended side effects.

Existing tests, 20 repetitions of `TestClean*`, full race suite, and conformance all pass; they do not cover the fresh-acquire commit window above. Temporary probe removed; review worktree is clean.

AlexAvilov:
Fixed at 218f417, both findings:

**BLOCKER**: confirmed the exact mechanism — `AcquireServeLease`'s fresh-acquire path ran on an unlocked "fast path" (`internal/loop/lease.go`), so it never contended with an external caller's own `WithAgentLock(id)`. `clean --mailbox`'s recheck+RemoveAll critical section couldn't exclude a concurrent boot's fresh lease acquisition at all — only the registration writer (which does share the lock) was actually protected.

Fix: unified the fresh-acquire link attempt and the stale-reclaim CAS under one `WithAgentLock(id)` critical section in `AcquireServeLease`. `link-no-clobber` was already atomic on its own (that guarantee never depended on the lock) — the lock's only real job was giving *external* holders of `WithAgentLock(id)` true exclusion, which the unlocked path never provided. All existing lease concurrency tests (single-winner, concurrent-reclaim, serialized-under-lock) pass unchanged under `-race`, confirming this changes no observable behavior. New `TestAcquireServeLeaseFreshAcquireExcludedByExternalLock` proves the fix directly — verified it fails against the prior code before confirming it passes with the fix.

**SHOULD-FIX**: `--owed` plan mode now does a plain lock-free read instead of entering `WithAgentLock` (whose `ensurePrivateDir` side effect was creating `state/<id>/` even though nothing was written). Safe because `SaveOwedLedger`'s atomic tmp+rename means a lock-free read can never observe a torn file — the only cost is the plan being possibly stale by the time `--yes` runs, which `--yes`'s own fresh locked read already accounts for. New `TestCleanOwedPlanIsLockFree` verified to fail against the prior always-locked code.

Full suite + conformance + `go vet` + `gofmt -l` + `go test -race` (including repeated runs on the lease/clean tests) all clean.

AlexAvilov:
claude-code round-3 delta verdict: LGTM at 218f417, with one scope note. The lease.go unification is the correct fix shape: the unlocked fresh-acquire path's atomicity never came from the lock (link-no-clobber owns that), so moving it inside withAgentLock(id) costs one uncontended acquisition at serve startup and exists precisely to give external critical sections (clean's) true exclusion — the comment now documents this. True PR diff vs main confirmed: clean.go/clean_test.go/dispatch.go/lease.go/lease_test.go only (the serve.go noise in the round-over-round diff is #82 arriving via main). Plan-mode --owed is now lock-free, answering the purity nit. SCOPE NOTE for the record: this PR now modifies substrate lease code beyond the A4 slice text — justified by the blocker, and the v2.5 plan keeps the launch-time lease (B7), so this unification is load-bearing going forward; codex as lease/fence domain reviewer has the final word on it. Merge on codex's round-3 approve + green CI, then #87 immediately after per the ordering gate.

AlexAvilov:
VERDICT: APPROVE

Reviewed exact base `0a2f994f1c86dd3eece92e803397bafc602922b0` / head `218f41708ea700dd17fcf30fa14146c5c2d02e84`.

- Exclusion gap closed: the fresh link and EEXIST/reclaim path now share one `withAgentLock` critical section in `internal/loop/lease.go:184-250`, composing correctly with the mailbox recheck/removal lock.
- Non-reentrancy audit passed: the lease closure calls no lock-taking function; the sole production caller at `internal/cli/serve.go:513` is outside an agent lock; test hooks enter acquisitions only from separate goroutines.
- EEXIST/release behavior and temp-inode lifecycle remain safe: the unique temp inode is prepared before locking, removed on every return, and a successful hard link leaves the claim inode intact.
- Plan purity fixed: `clean --owed` without `--yes` uses the lock-free read path at `internal/cli/clean.go:108-139`; the untouched-id test proves it creates no state directory.

Verification passed:
- fresh-acquire exclusion: count 100; race count 25
- owed-plan purity: count 100; race count 25
- focused lease race suite: count 20
- clean race suite: count 10
- `go test -race ./...`
- conformance `go test ./...`
- PR CI: Ubuntu, macOS, installer all green

Non-blocking documentation correction: `internal/cli/clean.go:249-260` still says the fresh-acquire path is intentionally unlocked and cannot be excluded by the target lock. That now contradicts the load-bearing round-3 fix and should be updated before or with merge.

PR #85, opened 2026-07-30
fix(test): make PTY startup size check deterministic

AlexAvilov:
## Summary

- replace the flaky short-lived stty output probe with a direct read of the child PTY shared kernel window state
- keep the child alive during the assertion and change no product code

## Verification

- focused test: 1,000 consecutive passes
- focused race test: 100 consecutive passes
- gofmt -l .
- go vet ./...
- go test ./...
- go test ./... -race
- go build ./...
- sh tests/install_test.sh
- conformance go test ./...
- shellcheck -s sh install.sh tests/install_test.sh
- goreleaser check

### DIFF
--- a/internal/runner/pty/pty_size_test.go
+++ b/internal/runner/pty/pty_size_test.go
@@ -2,9 +2,7 @@ package pty
 
 import (
 	"os/exec"
-	"strings"
 	"testing"
-	"time"
 
 	creackpty "github.com/creack/pty"
 )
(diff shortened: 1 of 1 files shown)


AlexAvilov:
VERDICT: APPROVE

Verified against base 7ba82d5.../head b907f09d... (matches the review ask exactly).

- No product code changed — the diff touches only `internal/runner/pty/pty_size_test.go`; `pty.go`'s `StartInheritSize` is untouched. Confirmed via `gh api .../pulls/85/files`.
- The invariant still holds: window size is genuinely a kernel-level property of the pty device, readable identically from either the master (`ptmx`, returned here) or the slave (attached to the child) via TIOCGWINSZ — it is not master-specific state. `StartInheritSize` calls `creackpty.StartWithSize(cmd, sz)`, and per the *removed* comment's own claim (which this PR doesn't dispute and I have no reason to doubt, since it describes creack/pty's own internals, unchanged here): `Setsize` happens before `cmd.Start()` forks the child, so the size is correct from before the child process exists — there is no window where the child could observe a stale/0x0 size. Reading `GetsizeFull(ptmx)` any time afterward is therefore equivalent to what the child would see on its own first read, and since this isolated unit test never wires up a resize loop, nothing else can change it in between. Same underlying invariant as before, proven directly instead of by racing a spawned `stty size` subprocess's stdout output — which is the actual, CI-scheduling-dependent flakiness source the old 500ms-poll comment was compensating for.
- Cleanup is correct: `sleep 10` keeps the child alive for the duration of the (now-instant) assertion instead of racing its natural exit; the `defer` kills and reaps it (`Kill` + `Wait`) so the test never leaves an orphan or a 10s hang. Errors from `Kill`/`Wait` are intentionally ignored (cleanup-only, not correctness-bearing).
- No portability concern: `Setsize`/`GetsizeFull`/`StartWithSize` are all pre-existing creack/pty API surface already used elsewhere in this same file — nothing new introduced for macOS vs. Linux.

Second reviewer, concurring with claude-code's APPROVE.

AlexAvilov:
Stability gate met at b907f09d4900bd049f0b78b9da64d49db22a0f01.

- macOS initial success: run 30584484673, job 91012524368
- required rerun success: same run attempt 2, job 91012894067
- additional full rerun success: attempt 3, job 91013156638
- claude-code-agentchute: APPROVE
- sonnet: APPROVE (https://github.com/agentchute/agentchute/pull/85#issuecomment-5136619131)

No product code changed.

PR #86, opened 2026-07-30
fix(send): preflight order, spool boundary, no recipient coaching (v2.5 plan A5)

AlexAvilov:
## Summary

- run sender and recipient preflights before consuming stdin
- remove recipient-registration coaching from send failures
- preserve raw bodies with semantic retry flags on pre-link failures
- treat successful link as the commit point and report post-link failures as exit-zero partial success

## Verification

- all named A5 regression tests
- gofmt -l .
- go vet ./...
- go test ./...
- go test ./... -race
- go build ./...
- sh tests/install_test.sh
- conformance go test ./...
- shellcheck -s sh install.sh tests/install_test.sh
- goreleaser check

## Scope

Implements v2.5 plan A5 only. B3 freshness semantics are intentionally excluded.


AlexAvilov:
VERDICT: APPROVE

Verified against base 7ba82d5.../head 0fdd7f21af... (matches the review ask). Checked the diff line-by-line against plan slice A5 + C28/C29, and confirmed the exact ordering/text/commit-boundary claims rather than trusting the PR description.

- **Preflight order**: sender-registered check and recipient inbox-dir preflight are both relocated before the stdin read, in the exact order the slice specifies (flags → ids → Discover → resolve from → sender check → recipient preflight → stdin → compose → deliver). `TestSendPreflightRunsBeforeStdin` proves it directly (asserts the stdin fd's read offset stays 0 on an unknown-recipient failure) rather than just asserting on error text.
- **Coaching removed**: the new `unknownRecipientError` text is a verbatim match for A5's specified (age-free, pre-B3) C29(a) wording — `unknown agent %q: no inbox/registration. Check the id (agentchute status).` — and is used at both remaining sites (the stat-based preflight, and the `os.IsNotExist` bubble in `preserveSendBody` for the recipient-disappears-mid-send race). `send_test.go`/`send_gate4_test.go` updated to assert the new text and explicitly assert "register" never appears. The sender-side self-registration coaching (boot yourself) is untouched and is a different thing (not the impersonation-door concern C29 is about) — correctly left alone.
- **Spool boundary is genuinely pre-link only**: `writeSeqMessageWithCommit`/`SendSeqMessageWithCommit` correctly report `committed=true` the instant the canonical link succeeds — including when the post-link `syncDir` barrier then fails — so `cmdSend`'s `if sendErr != nil && !committed` branch only spools true pre-link failures. Verified with a forced-failure test seam at both layers (`internal/loop/seq_test.go`'s `TestSendSeqMessageWithCommitReportsPostLinkSyncFailure`, and `send_a5_test.go`'s `TestSendPostLinkSyncFailureIsPartialSuccess`), not just asserted.
- **Spool content is the raw body**, captured before the `--ask` heading is applied and before `ComposeMessage` — matches C28's corrected intent (raw body, not the fully-composed file), and avoids a double `## ASK` heading on retry. `sendRetryCommand` reconstructs `--ask`/`--reply-by`/`--reply-to` from the *original* argv (via `fs.Visit` + `originalFlagValue`, not parsed/defaulted values), so a retry can't subtly diverge from what the user actually typed — verified end-to-end by `TestSendSpoolRetryPreservesAskSemantics` and `TestSendSpoolRetryPreservesReplyTo`, which replay the spool and check the delivered message's wire shape and the resulting `.owed` entry, not just the retry-line text.
- **Both partial-success paths exit 0** and never spool: post-link sync failure (`TestSendPostLinkSyncFailureIsPartialSuccess`) and post-link `RecordOwed` failure (`TestSendPartialSuccessOnOwedFailure`) — both assert `err == nil`, the exact `WARNING: ... Do NOT resend.` text, zero spool files created, and that the message really did land in the recipient's inbox.
- **No B3 scope creep**: the recipient preflight is still a plain `os.Stat` on the inbox directory — no registration-row read, no freshness/staleness comparison. Confirmed this is the only recipient-liveness check anywhere in the diff.
- Spool mechanics (0700 dir, 0600 file via both `O_CREATE` mode and an explicit `Chmod`, `O_EXCL` so a same-microsecond collision fails closed instead of silently overwriting a different pending body, cleanup-on-partial-write via a `cleanup` flag in a defer) all check out — same defensive tmp-file discipline already used elsewhere in this codebase (`atomicWriteFile`/`writeAndSyncOpenFile`).

No blocking findings. Nit (non-blocking): the sender's own `last_seen` update now runs off a fresh `time.Now().UTC()` call taken before stdin is read, rather than the `now` computed later for message composition — two timestamps where there used to be one. Harmless in practice (stdin read is near-instant for a piped body) and arguably more correct (measures "sender was active" at the true call time), not something I'd hold the PR for.

Second reviewer, concurring with claude-code's LGTM.

AlexAvilov:
Current gate status at 0fdd7f21af92a5ac616932f651344edd436d4b12:

- Sonnet: APPROVE (https://github.com/agentchute/agentchute/pull/86#issuecomment-5136642736)
- Claude review: pending
- Ubuntu CI: success
- installer CI: success
- macOS CI: failed only in the pre-A5 baseline test internal/runner/pty/pty_size_test.go:68 with the known empty stty-output harness failure

The prerequisite fix is PR #85, which has both required approvals and three successful macOS executions including the required rerun. After #85 lands, this PR needs a macOS rerun against the fixed base.

AlexAvilov:
claude-code review verdict: needs-changes. BLOCKER (probed on the PR tree): the retry line rebuilds --reply-by via a raw-argv rescan (originalFlagValue, send.go:42) instead of the parsed value — '-reply-by 45m' (legal single-dash form) yields an empty value so the advertised retry command fails to parse, and any later argv token beginning '--reply-by=' (e.g. inside --body) clobbers the real deadline so the retry silently delivers a WRONG reply-by. This is the exact silent ask-corruption C28 exists to prevent. Fix: emit parsed replyBy.String(), as the --reply-to reconstruction already does. SHOULD-FIX (probed): --ask + post-link sync failure returns partial success exit 0 but skips RecordOwed silently — reply_required lands with no obligation, the 'silent leak' send.go's own comment names; either attempt RecordOwed after the sync-failure partial success (bookkeeping is independent of the dir-sync barrier) or name the skipped obligation in the warning. macOS red is the pre-#85 harness — #85 is now MERGED, so a rebase/rerun clears it. Disclosure: my test-audit half was cut short by a resource limit mid-review; the findings above are empirically probed, and the fix-round re-review will cover the full test audit.

AlexAvilov:
Final fix-round proof at head a14d1ada988f9ba7d0d6e3397031fe48e56db7f3:

- Retry reconstruction now uses the parsed reply-by duration; the regression covers legal single-dash input plus a body token that previously clobbered the raw argv rescan.
- Committed post-link sync failures now continue through RecordOwed for --ask sends; the regression asserts delivery, warning, no spool, reply_required, and the 45m owed entry.
- Local: gofmt, vet, test, race, build, conformance all green; focused regressions passed 100 regular and 10 race-enabled repetitions.
- CI run 30588591061: Ubuntu, macOS, and installer jobs green.
- Claude exact-head delta verdict: APPROVE (agentchute seq132); Sonnet prior verdict: APPROVE.

Merged by the coordinator as ca837244173cd36e77c3e035bcc85ceea892b8a4.

PR #87, opened 2026-07-30
feat(check): age banner + stale-owed prune offer (v2.5 plan A3)

AlexAvilov:
## Summary
- `check` announces a message's age above its body once it's older than 24h — `[!] this message is N days old (sent YYYY-MM-DD)` — in both the normal claim path and the crash-residue redelivery path. Age is sourced from file mtime today via a small `messageAge(msg, now)` helper (plan B7 will later swap in the filename timestamp for new-format names, in this one place, per the slice text).
- `check` surfaces this agent's own **expired** reply obligations with a copy-pasteable prune command — `stale reply obligation (<ref>, expired <age> ago) — prune with: agentchute clean --owed --as <id>` — print-only, never auto-removes. A corrupt/unreadable `.owed` ledger is a warning, not a check failure, mirroring `gate.go`'s existing tolerance for the identical condition.

This implements slice A3 of `docs/decisions/agentchute-v2.5-implementation-plan.md` (rev 2.3) — team-approved, brief-by-reference. No wire/protocol change; Group A, independently releasable.

**Coupling with A4** (noted per the plan): the prune-offer hint text names `agentchute clean --owed --as <id>`, the command A4 (#84) adds. This PR builds/tests fine on its own (the hint is plain text, not a function call) but both should land before either merges, so a released binary's hint never names a not-yet-existing command.

## Deviation from the slice text (disclosed)
The stale-owed prune offer is placed **ahead of** the `(inbox empty)` early return in `cmdCheck`, not after it. The slice's cited anchor point (`internal/cli/check.go`, "after :226 note, before UpdateLastActive :231") sits *behind* that early return — literally there, the offer would never print whenever the inbox has nothing to claim this turn, which is exactly the situation where a returning agent is most likely to want to see a stale obligation from weeks ago. The offer is read-only (no mutation) and unconditional on `--no-archive`/`claimed`, so moving it earlier has no other behavioral cost.

## Test plan
- [x] `TestCheckPrintsAgeBannerOnOldMail` — 3-day-old mail prints the banner
- [x] `TestCheckNoBannerOnFreshMail` — fresh mail prints no banner
- [x] `TestCheckOffersStaleOwedPrune` — expired obligation surfaced with the exact prune command, alongside `(inbox empty)`
- [x] `TestCheckDoesNotOfferOutstandingObligation` — a not-yet-due obligation is never offered
- [x] `TestCheckCorruptOwedLedgerWarnsNotFatal` — corrupt ledger warns on stderr, `check` still exits 0
- [x] `internal/cli/n3_sanitize_test.go` — unaffected (banner is program-generated text, printed outside the sanitized-body path), still green
- [x] `internal/cli/consume_boundary_test.go` — untouched, still green
- [x] Added `mustWriteAgedInbox` test helper (cited by the plan at a line number that doesn't exist in the current tree — added it following the existing `os.Chtimes` back-dating pattern already used in `doctor_test.go`/`inbox_test.go`)
- [x] Full suite: `go test ./...` (root module; no conformance-module surface touched), `go test -race ./internal/cli/...`, `go vet ./...`, `gofmt -l` — all clean


AlexAvilov:
claude-code review verdict: needs-changes, three should-fixes (two proven by mutation testing). 1. The prune offer reads the ledger BEFORE the claim loop, so a turn that consumes the very reply discharging an expired obligation still advertises it (reproduced: offer printed, then the discharging reply, ledger empty afterward). Move the offer computation AFTER the claim loop and make the empty-inbox branch fall through instead of returning — this keeps your (correct) empty-inbox rationale, fixes the misleading print, and fixes the redelivered/fresh-mail sandwich ordering. 2. The 24h banner boundary is unpinned: setting the constant to 1h leaves the whole suite green; add an aged-but-under-threshold case (e.g. 23h, assert no banner). 3. The print-only covenant is unpinned: adding an auto-ClearOwed inside the offer loop leaves the suite green; reload the ledger after cmdCheck and assert the expired entry survives. NITS: assert the banner prints above the ---- header (ordering, not just Contains); add banner coverage on the redelivery and --no-archive paths (all three share printConsumedBody today, but a refactor would lose it silently). Verified clean otherwise: banner in all three display paths outside the sanitizer, offer read-only, hint text matches #84's real flags, corrupt-ledger tolerance mirrors gate.go, messageAge is the single dual-read seam.

AlexAvilov:
Fixed at 5f4c400, all three should-fixes plus both nits, each verified against the reviewer's exact failure mode:

**Should-fix (same-turn discharge)**: reproduced it first (offer printed, then the discharging reply, ledger empty afterward — exactly as described), then moved the offer computation to after the claim loop and changed the empty-inbox branch to fall through instead of returning, preserving the original A3 deviation (offer still surfaces on an empty inbox). New `TestCheckPruneOfferReflectsSameTurnDischarge` fails against the pre-fix ordering and passes now.

**Should-fix (banner threshold unpinned)**: added a 23h-aged case. Confirmed it fails if `oldMailBannerAfter` is shrunk to 1h (the existing near-zero-age "fresh mail" test doesn't catch that, since 0 < any plausible threshold).

**Should-fix (print-only covenant unpinned)**: `TestCheckOffersStaleOwedPrune` now reloads the ledger after `cmdCheck` and asserts the expired entry survives. Confirmed it fails if an auto-`ClearOwed` is injected into the offer loop.

**Nits**: the banner assertion now checks the substring's *position* (must be before the `----` header), not just presence; added coverage for the redelivery path (crash-orphaned `.claimed` residue) and the `--no-archive` dry-run path — all three `printConsumedBody` callers now individually pinned.

Full suite + `go vet` + `gofmt -l` + `go test -race` clean.

AlexAvilov:
claude-code delta verdict: LGTM at 2718651. Verified in code: offer computation moved after the claim loop reading the post-discharge ledger, empty-inbox branch falls through; 23h under-threshold case pins the 24h boundary; post-check ledger reload pins the print-only covenant; banner ordering asserted; redelivery + --no-archive coverage added. Merge waits on codex's independent cross-review.

AlexAvilov:
VERDICT: APPROVE

Reviewed exact base `ca837244173cd36e77c3e035bcc85ceea892b8a4` / head `271865101001a8bfb30d9e05735c23fb3e90ebfe`.

- C18 is implemented through one `messageAge` seam with the exact >24h threshold, banner-before-header ordering, UTC sent date, and shared fresh/redelivery/`--no-archive` display path (`internal/cli/check.go:15-25`, `:308-327`).
- C19 runs after the claim loop, so a same-turn reply clears its obligation before the offer is computed; the empty-inbox path falls through, only expired entries print, and corrupt ledgers warn without failing (`internal/cli/check.go:164-171`, `:254-271`).
- Tests pin the 24h boundary, banner ordering and all display paths, print-only behavior, expired-only filtering, and same-turn discharge (`internal/cli/check_age_owed_test.go:20-217`).

Verification: focused A3 tests passed 25 repetitions; full `go test -race ./...` and conformance module passed; GitHub CI is green. No blocking findings.

Merge-order gate: A3 prints `agentchute clean --owed`, so the plan requires A4 to land first or in the same release. PR #84 is currently changes-requested; do not merge #87 ahead of the corrected A4 command.

PR #88, opened 2026-07-30
feat(check): delete corrective auto-send, keep quarantine (v2.5 plan A6)

AlexAvilov:
## Summary
Malformed inbox mail is still quarantined with a local `stderr` warning; no machine mail is manufactured about it anymore. This removes the filename-text laundering channel decision §9 calls out: the corrective's `malformedItem` path (attacker-influenced text) was being carried into an authentic, sequenced message sent from a trusted peer's identity — enabling mutual finish-gate flooding, with its only real-world observed output being machine mail complaining about malformed machine mail.

- `internal/loop/message.go`: deleted `CorrectiveBody`, `SendCorrective`, `correctiveIdempotencyKey`.
- `internal/cli/check.go`: both quarantine call sites (malformed §6.1 filename, malformed §6.4 frontmatter) keep `QuarantineInboxFile` + the local warning; deleted the `SendCorrective` calls and the offender-inference plumbing that fed them.
- `internal/loop/inbox.go`: deleted `InferSenderFromFrontmatter` (its sole caller was the deleted notify path). Kept `InferSenderFromFilename` — now caller-less in production — with a comment noting it's retained for B6's dual-grammar extension, per the plan.
- `AGENTCHUTE.md` §11.1: deleted the "Notify offender" enforcement step and renumbered the remaining two (this is an embedded-spec edit — `//go:embed` means it changes the binary, not just docs).
- Carry-along fix (nit from #84's review, explicitly asked to land in whichever of A6/A7+A8 goes first): corrected a stale comment in `clean.go` that still described `AcquireServeLease`'s fresh-acquire path as unlocked — the round-3 fix already merged on #84 unified it under one lock.

This implements slice A6 of `docs/decisions/agentchute-v2.5-implementation-plan.md` (on main) — team-approved, brief-by-reference. No wire/protocol change; Group A, independently releasable.

## Test changes
- Deleted the four `SendCorrective`/`CorrectiveBody` unit tests and `TestInferSenderFromFrontmatterReadsFromField`, per the slice's own test list.
- `frontmatter_characterization_test.go` (three-parser comparison) trimmed to the two parsers (B: `parseFrontmatter`, C: `ParseMessageFrontmatter`) that survive A6 — full consolidation to one parser is B8's job, noted as such in the file's own doc comment so it isn't mistaken for scope creep here.
- Added two new tests pinning A6's own done-when directly (not just "the deleted function doesn't exist"): `TestCheckQuarantinesMalformedFilenameWithoutCorrective` and `TestCheckQuarantinesMalformedFrontmatterWithoutCorrective` — each seeds mail that looks like it came from a real, registered peer, runs `check`, and asserts quarantine + a local warning fire while that peer's inbox receives nothing.

## Deviations
None from the slice text. The carry-along comment fix in `clean.go` isn't part of A6's own scope but was explicitly requested to land in whichever of A6/A7+A8 opens first.

## Verification
- [x] `grep -rn "SendCorrective\|CorrectiveBody" --include=*.go .` — empty (including comments, not just code)
- [x] `TestQuarantineInboxFile`, `TestGateBlocksOnMalformedInbox` — kept green, unmodified
- [x] Full suite: `go test ./...` (root + conformance modules), `go test -race ./internal/cli/... ./internal/loop/...`, `go vet ./...`, `gofmt -l` — all clean

### DIFF
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -256,8 +256,7 @@ Every agent participates in keeping the pool healthy.
 ### 11.1 Enforcement action
 Triggers include malformed inbox filenames, unparseable frontmatter, or unparseable peer registrations.
 1. **Quarantine**: atomic move to `.agentchute/loop/malformed/`.
-2. **Notify offender**: send a corrective message to the inferred sender (the correction is plain body text; `task`/`status` are no longer normative wire fields). Sender inference order: filename capture → frontmatter `from:` → no notify.
-3. **Continue**: do NOT block the sender or the turn.
+2. **Continue**: do NOT block the sender or the turn.
 
 A well-formed canonical seq file is never quarantined; only a genuinely-unrecognized name is enforced on.
 
--- a/internal/cli/check.go
+++ b/internal/cli/check.go
@@ -109,27 +109,6 @@ func cmdCheck(args []string) error {
 			}
 			fmt.Fprintf(os.Stderr, "warning: quarantined %s (malformed §6.1 filename) -> %s\n",
 				name, quarantined)
-
-			// Sender inference order per §11.1: filename capture → frontmatter
-			// from: → no notify.
-			offender, ok := loop.InferSenderFromFilename(name)
-			if !ok {
-				offender, ok = loop.InferSenderFromFrontmatter(quarantined)
-			}
-			if !ok {
-				fmt.Fprintf(os.Stderr, "  offender unidentifiable; corrective notify skipped\n")
-				continue
-			}
-			if offender == agentID {
-				fmt.Fprintf(os.Stderr, "  inferred offender is self; corrective notify skipped\n")
-				continue
-			}
-			if _, err := loop.SendCorrective(cfg, agentID, offender,
-				quarantined, "filename does not match §6.1", "§6.1"); err != nil {
-				fmt.Fprintf(os.Stderr, "  corrective send to %s failed: %v\n", offender, err)
-				continue
-			}
-			fmt.Fprintf(os.Stderr, "  notified %s\n", offender)
 		}
 	} else if len(skipped) > 0 {
 		fmt.Fprintf(os.Stderr, "warning: %d non-§6.1 file(s) in inbox; --no-archive suppressed §11 enforcement:\n", len(skipped))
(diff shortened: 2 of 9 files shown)


AlexAvilov:
claude-code review verdict: LGTM pending green CI. Verified: SendCorrective/CorrectiveBody/correctiveIdempotencyKey and both check.go call sites deleted; quarantine + local warning retained with two new pinning tests asserting no outbound mail on either malformed path; InferSenderFromFrontmatter deleted; InferSenderFromFilename retained caller-less with the exact retained-for-B6 comment the plan specifies; AGENTCHUTE.md §11.1 notify-offender step deleted (embedded-spec edit — flagging per the embed rule that this changes the binary; no hook/doctor surface touched); the clean.go stale-comment nit rode along as agreed. Remaining grep hits are test-comment documentation of the deletion, which is fine. Merge on codex's cross-review + green CI.

AlexAvilov:
VERDICT: APPROVE

Reviewed exact base `caa2f3421c2405c378be16e43ef9f48898abc57b` / head `4d6217f39a6afc70b99da3b52cae15c825999265`.

No blocking findings.

- Deletion inventory is complete: `CorrectiveBody`, `SendCorrective`, `correctiveIdempotencyKey`, `InferSenderFromFrontmatter`, and both production call sites are gone. Remaining symbol-name hits are test/history comments; no quarantine path reaches `SendSeqMessage` or another outbound-mail producer.
- Security rationale is realized: malformed-name and malformed-frontmatter branches now end after `QuarantineInboxFile` plus the local warning. Attacker-influenced quarantine path text is never composed into trusted sequenced mail.
- Quarantine behavior is unchanged minus the send: `QuarantineInboxFile` is byte-unchanged; the `check.go` delta only deletes offender inference/send/output blocks, preserving move, warning, failure, dry-run, and claimed-count behavior.
- Both new A6 tests use registered `bob` with a writable inbox and assert it remains empty. Reintroducing either former corrective call would land mail and fail the corresponding test; they also reject the old `notified` output.
- `AGENTCHUTE.md` changes only §11.1's notify-offender step plus the required Continue renumber. The embed changes binary bytes as expected; doctor/hook implementation and templates are untouched, and their suites pass.
- The carry-along `clean.go` comment now accurately states that fresh acquire and reclaim share the target lock.

Verification:
- A6 tests: count 100; race count 25
- quarantine invariant: race count 50
- doctor/assets/hooks race suite
- `gofmt -l .`, `go vet ./...`, `go build ./...`
- `go test -race ./...`
- conformance `go test -race ./...`
- PR CI: Ubuntu, macOS, installer all green

Non-blocking cleanup: production-facing text still promises deleted behavior at `internal/cli/pending.go:206,231` and `internal/cli/gate.go:320,421` ("quarantine + notify"), while `internal/cli/check.go:39` still advertises "corrective sends." Update those strings to quarantine-only; nearby stale comments in `check.go:76,96-101,163-166` and `internal/loop/inbox.go:168-170` should be corrected with them.

PR #90, opened 2026-07-31
feat(loop): dual-read old and new message identities (v2.5 plan B6)

AlexAvilov:
## Summary
- add strict timestamp identity filename, reference, and stamp parsers
- dual-read old and timestamp formats in inbox, claimed residue, reply refs, and owed obligations
- preserve the legacy seq writer; no B7 writer or allocator changes

## Verification
- gofmt -w .
- go vet ./...
- go test ./...
- go build ./...
- go test -race ./...
- cd conformance && go test ./...
- sh tests/install_test.sh

## Deviations
- None.

### DIFF
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -151,6 +151,8 @@ The committed identity is the full delivery key `(to, from, seq)`:
 
 The reference encoding is the canonical filename `from-<from>_seq-<020d>.md` (`seq` zero-padded to 20 digits). A plain lexicographic sort of one sender's files is therefore **exact per-sender FIFO with no clock**. Cross-sender order is **advisory arrival order** (non-normative) — the protocol does not promise a global total order (a real total order would need a freshness CAS the mount can't cheaply give, and is unneeded).
 
+During the v2.5 dual-read migration, the reference implementation preserves per-sender FIFO for the one-way, pool-at-once cutover by sorting legacy seq messages before timestamp-format messages, then sorting each grammar by its own monotonic key. Known migration-window limitation: if a pool writes timestamp-format messages, rolls back and writes new legacy messages, then rolls forward again, the old-first grammar rank can place those rollback writes before earlier timestamp writes. Using file mtime as a cross-grammar key is worse because copying and archiving can perturb it; no global ordering machinery is added for this two-transition edge case.
+
 `seq` is **write-ahead durable**: the counter is committed *before* the message links, so a crash can only ever produce a GAP (an allocated seq whose message never landed), never a reuse for different content. Gaps are legal — `seq` is identity + sort key, not a no-gap contract.
 
 The canonical `from-<from>_seq-<020d>.md` is the only inbox filename format. A name that does not parse as a canonical seq filename is unrecognized: it is skipped by the lister and quarantined by `check` (§11.1).
--- a/internal/cli/check.go
+++ b/internal/cli/check.go
@@ -309,6 +309,10 @@ func displayConsumed(cfg *loop.Config, agentID string, msg loop.Message, content
 			if err := loop.ClearOwed(cfg, agentID, key); err != nil {
 				fmt.Fprintf(os.Stderr, "warning: failed to clear owed obligation %s: %v\n", ref, err)
 			}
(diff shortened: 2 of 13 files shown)


AlexAvilov:
claude-code review verdict: **needs-changes**, plus a PLAN AMENDMENT — my two reviewers independently found the same defect and it is a spec-level issue the plan missed, not a coding error in this PR.

**THE DEFECT (both reviewers, independently, with probes).** Admitting both grammars into one lexicographically-sorted list breaks per-sender FIFO — which is a NORMATIVE conformance guarantee (O1, CONFORMANCE.md:44, AGENTCHUTE.md:152, TestO1_PerSenderFIFO). New-format names start with a digit (0x3n), old-format with 'f' (0x66), so for the SAME sender every new-format message sorts before every old-format one. Probed end-to-end: an inbox holding bob's seq-1, seq-2 and one new-format bob message lists the new one FIRST, and `check` displayed the newer body before the older. B6 alone is inert (no writer yet), but this is precisely the §3 cutover state — an old backlog sitting in inboxes while v2.5 senders write the new format — and B7 inherits it. The sort comment at inbox.go:182-185 still asserts the guarantee that is now false, and the PR's new mixed-directory test collects into a map, so it pins no order at all.

**RULING — amend the plan, fix the sort (do not weaken O1).** The plan's C3 rationale ("timestamp-first so plain sort = chronological") is correct WITHIN a grammar and silently wrong ACROSS grammars; that gap is mine, from the planning round. The fix that preserves the guarantee: sort by (sender, grammar-rank, own-key) with OLD-FORMAT RANKING FIRST, then seq numerically for old and stamp lexicographically for new. This is correct for the forward migration by construction: a given binary writes exactly one grammar and the cutover is pool-at-once, so within one sender every old-format message was written before every new-format one. O1 holds through the migration window, unweakened.

One honest caveat to document, not to engineer around: plan §3 step 8 deliberately keeps rollback available, and a rollback-then-roll-forward could produce an old-format message written AFTER new-format ones, which this rank rule would misorder. That is a two-transition edge case, the consumer is an agent reading the whole batch, and the alternative (mtime as a cross-grammar key) is worse — archiving and copying perturb mtime. Record it in the spec's ordering section as a known migration-window limitation; do not add machinery for it.

**REQUIRED IN THIS PR:** the ranked sort (both listers — the claimed lister reuses the same sort at inbox.go:406-408), a test pinning mixed-directory ORDER for one sender (not a map), and the corrected sort comment.

**ALSO REQUIRED — comment rot, A1 class.** This project shipped a whole slice to delete comments that lie; five now contradict the code in the same file they document: check.go:20-23 (messageAge says the timestamp grammar "doesn't exist yet" and that it is "the one place B6 edits" — B6 landed and correctly edited parseAnyInboxName instead; probe shows a 5-day-old stamp with a fresh mtime prints "5 days old", so the comment actively misdirects the B7 implementer); inbox.go:23 ("file mtime") contradicting inbox.go:16-18 rewritten in this same diff; inbox.go:102 and :165 still seq-only and still citing the §11 corrective that A6 deleted one commit earlier; pending.go:235 "the canonical (to,from,seq) ref" though owedLine now renders either form; owed.go:15 likewise.

**ONE BEHAVIOR CHANGE TO REVERT OR JUSTIFY.** ClearOwed gained key validation the base never had, so a peer-crafted old-format in_reply_to now emits a stderr warning where the base was silent — an A/B of base vs head on an identical fixture differs by exactly that line. No state change, exit code unchanged, but this slice's contract is "no behavior change for existing traffic", and the validation is not in the B6 spec (which names only Key()/MatchesRef for ClearOwed; RecordOwed already validates on write). Remove it, or keep it with a test and say why it belongs here.

**NIT worth a test, not a fix:** ParseTsFilename leaves To empty (correct — it is not encoded in the name) and validateOwedKey rejects an empty To, so a TsID taken straight from a filename can never be recorded as an obligation. That is a good loud failure and a live footgun for B7, whose RecordOwed call site must set To explicitly. Pin it with a test here.

Otherwise the dual-read core is right: the single choke point covers both listers, format-matching reply refs work, the anti-spoof triple survives in both grammars, neither-form owed entries reject at load, garbage still quarantines, and nothing writes the new format yet.

PR #91, opened 2026-07-31
feat(registration): soft-state registration + lazy sweep (v2.5 plan B1)

AlexAvilov:
## Summary

Implements v2.5 plan slice B1 (`docs/decisions/agentchute-v2.5-implementation-plan.md`, §2 B1 + constants C9–C13): registration becomes "reachable now-ish" instead of relying on `.live`/explicit deregistration.

- **C9** — `--stale-after` pool config flag (`setup.go`), `loop.StaleAfter(cfg)` reader (default 1h, tolerant of absent/corrupt state), `setup.json` writes converted to atomic temp+rename, `update` replays the flag on resync.
- **C13** — `loop.HeartbeatRegistration(cfg, template, leaseToken)` replaces `UpdateLastSeen`: lease-gated (empty token errors; `ErrFenced` on a reclaimed/absent claim, writing nothing), self-healing (recreates a swept/missing row from `template`), merges with existing on-disk `Body`/`WorkingRepos`/`LastActive`/launch provenance (template wins for everything else, empty-falls-back-to-existing for provenance — see round 2 item 4). Sole caller is `serve`'s poll tick. `check`/`send`/`status` no longer refresh liveness — their registration-exists preflights stay, with the existing "run `agentchute boot --as <id>`" hint.
- **C11/C12** — `loop.SweepStaleRegistrations(cfg, selfID, now)`: candidates are rows past `StaleAfter` with an absent/stale serve claim (never self); capped at 5/pass; corrupt rows use file mtime as an age proxy; README/`*.example.md` are never touched; each candidate is re-checked (age + lease) under `WithAgentLock` immediately before delete, closing the scan→delete TOCTOU window. Triggered from `boot` (after self-register, before the inbox peek) and `serve`'s poll tick (10-minute cadence, immediate on the very first tick).
- **Quitting agent does nothing** — `markRunnerOffline` and its shutdown-defer call are deleted; a dead lane just stops heartbeating and ages out. `update.go`'s `activeAgentIDs` is now interim age-based (row age ≤ `StaleAfter`) instead of `Status != offline`.

**Done-when** (proven in `internal/cli/b1_convergence_test.go`, `TestB1ConvergencePoolWithDeadRowsAndLiveServe`): one live serve + two dead rows — dead rows gone after a single poll tick; the live row survives across ticks; killing the serve (release lease + backdate its row) and running a *peer's* boot sweeps it; restarting (booting again) recreates it.

## Deviations from plan text (disclosed per team convention)

1. `filelock_test.go`'s `TestRunnerOfflineNotResurrectedByConcurrentLastSeen` was **deleted** rather than rewritten to a `HeartbeatRegistration` equivalent — it modeled a race against `markRunnerOffline`, which B1 removes entirely, so the scenario no longer exists in production. Replaced the adjacent `TestUpdateLastSeen_NoLostUpdateUnderConcurrency` with a `HeartbeatRegistration`-based twin (same torn-write/lost-update coverage, now lease-gated).
2. `HeartbeatRegistration`'s signature omits `now time.Time` (uses `time.Now().UTC()` internally) — the plan's literal C13 signature doesn't list it, though sibling functions (`UpdateLastActive`) do take it. Kept literal since no test needed injectable time at this call (age-threshold math lives entirely in `SweepStaleRegistrations`, which *does* take `now` per plan).
3. `setup.go`'s default `--stale-after` fallback serializes as `loop.DefaultStaleAfter.String()` = `"1h0m0s"` rather than the literal `"1h"` — functionally identical (`time.ParseDuration` round-trips either), tied to the single source of truth instead of a duplicated literal.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- [x] `go test ./...` and `go test -race ./...` green (env stripped of `AGENTCHUTE_*`)
- [x] Mutation-verified: sweep's never-self guard, sweep's under-lock re-check, boot's sweep call (twice — unit + convergence integration), setup's stale-after preservation-on-resync — each reverted, confirmed the targeted test fails, restored, confirmed it passes again
- [x] Convergence Done-when test passing

Reviewers: claude-code, codex. Do not merge — merges on the gate per program convention.

## Review round 2 (claude-code panel + grok residuals) — head `bef9e53`

All items independently probe-reproduced against pre-fix code before touching anything, then revert-verified after (revert → confirm targeted test fails → restore → confirm it passes).

1. **BLOCKER, fixed** — stray non-agent-id `.md` in `agents/` (e.g. `NOTES.md`) no longer hard-errors the whole sweep pass. `SweepStaleRegistrations`'s scan loop now skips any basename failing `ValidateAgentID` before it ever reaches `sweepOneCandidate`'s `withAgentLock`. Test: `TestSweepStaleRegistrationsSkipsNonAgentIDFilenames`.
2. **BLOCKER, fixed** — the lease cross-check (`claimProvablyDead`) now fails CLOSED: only `os.ErrNotExist` counts as "no lease, safe to delete"; any other read error (corrupt claim, EACCES, symlink rejection, oversize) means "cannot prove dead" and the candidate is skipped, consistent with `VerifyFence`/`AcquireServeLease`'s existing fail-closed direction. Test: `TestSweepStaleRegistrationsFailsClosedOnCorruptClaim`.
3. **SHOULD-FIX, recorded not re-engineered** — `sweepOneCandidate`'s `withAgentLock` still creates `state/<id>/` + `.lock` as a side effect, for every candidate evaluated and for the id it deletes. Documented in `sweep.go`'s doc comment as the same accepted tradeoff `clean.go` (`cmdCleanOwed`, `--mailbox` apply path) already lives with: the lock is load-bearing for correctness (closes the TOCTOU window), cleanup would race concurrent activity for cosmetic benefit only, and doctor/gate/status never read these dirs.
4. **SHOULD-FIX, fixed** — `HeartbeatRegistration`'s merge now also preserves `LastActive` and launch provenance (`LaunchedBy`/`ShimName`/`HookEvent`, empty-template-value-falls-back-to-existing, matching `performRegister`'s own merge rule) instead of only `Body`/`WorkingRepos`. Tests: `TestHeartbeatRegistrationPreservesLastActive`, `TestHeartbeatRegistrationPreservesLaunchProvenanceTemplateDoesNotSet`.
5. **SHOULD-FIX, fixed** — added two tests decomposing the composite revival-hook coverage: `TestSweepStaleRegistrationsUnderLockRecheckSurvivesLeaseOnlyRevival` and `...SurvivesHeartbeatOnlyRevival`, each isolating one re-check clause (age vs. lease) as independently load-bearing, closing the "mutation-verified" gap the composite-only hook left.
6. **grok residual, fixed** — a future-dated `last_seen` (clock skew/hand edit) no longer grants permanent sweep immunity. `registrationAge` clamps any negative age to `math.MaxInt64` via `clampNonNegativeAge`, erring toward eventual cleanup rather than toward immortality (a registration row is discoverable display state, not a lock). Test: `TestSweepStaleRegistrationsFutureLastSeenIsNotImmortal`.
7. **grok residual, recorded not implemented** (cross-slice, correctly out of scope for B1) — `boot`/`register` advance `last_seen` without a lease, so a `boot --as ghost` loop with no runner behind it keeps a roster entry looking fresh indefinitely. Harmless under B1 (registration freshness gates nothing yet); flagged here for **B3 and B9**: once send-reachability or any other decision keys off registration age, a lease-less boot loop becomes a black-hole attractor that reads as reachable while nothing is actually polling the inbox. No code change in this PR.

CI green on `bef9e53` (build/vet/gofmt/test/test-race, `AGENTCHUTE_*` stripped).

### DIFF
--- a/internal/cli/b1_convergence_test.go
+++ b/internal/cli/b1_convergence_test.go
@@ -1,0 +1,108 @@
+package cli
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"github.com/agentchute/agentchute/internal/loop"
+)
+
+// TestB1ConvergencePoolWithDeadRowsAndLiveServe is B1's own Done-when
+// acceptance (v2.5 plan B1), proven in a test rather than by hand: a pool
+// with one live serve and two dead rows converges — dead rows are gone
+// after boot or a serve poll tick; the live row survives indefinitely;
+// killing the serve lets its own row age out and be swept by a peer's boot;
+// restarting recreates it.
+//
+// A very short --stale-after ("50ms") lets the test control staleness by
+// backdating registration timestamps rather than sleeping in real time; a
+// poll tick's lazy sweep is due immediately on its first call regardless
+// (lastSweep's zero value), so no sweepInterval shrinking is needed either.
+func TestB1ConvergencePoolWithDeadRowsAndLiveServe(t *testing.T) {
+	root := setupShortRunFixture(t)
+	cfg, err := loop.Discover(loop.DiscoverOpts{Cwd: root})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := writeSetupPoolState(cfg, "runner", nil, "50ms"); err != nil {
+		t.Fatal(err)
+	}
+
+	now := time.Now().UTC()
+	writeConvergenceRegistration(t, cfg, "dead-one", now.Add(-time.Hour))
+	writeConvergenceRegistration(t, cfg, "dead-two", now.Add(-time.Hour))
+
+	// The live serve: a real runnerRuntime (real lease, real registration).
+	rt := newPollTestRuntime(t, cfg, "live-agent")
+
+	// dead-one/dead-two gone after ONE poll tick (the "≤10min of serve" path;
+	// lastSweep's zero value makes the very first tick due immediately).
+	rt.pollOnce()
+	for _, id := range []string{"dead-one", "dead-two"} {
+		if _, err := os.Stat(cfg.AgentRegistrationPath(id)); !os.IsNotExist(err) {
+			t.Fatalf("%s should be swept by serve's poll tick: stat err = %v", id, err)
+		}
+	}
+	live, err := loop.ReadRegistration(cfg.AgentRegistrationPath("live-agent"))
+	if err != nil {
+		t.Fatalf("live-agent row missing after its own poll tick: %v", err)
+	}
+	// Registration timestamps round-trip at whole-second precision
+	// (formatTimestamp uses time.RFC3339), so compare against `now` truncated
+	// the same way rather than a spurious same-second sub-second mismatch.
+	if live.LastSeen.Before(now.Truncate(time.Second)) {
+		t.Fatalf("live-agent row was not heartbeated: LastSeen = %s", live.LastSeen)
+	}
+
+	// The live row survives across further ticks.
+	rt.pollOnce()
+	if _, err := os.Stat(cfg.AgentRegistrationPath("live-agent")); err != nil {
+		t.Fatalf("live-agent row should survive while heartbeating: %v", err)
+	}
+
+	// Kill the serve: release its lease (no more heartbeats), then let its
+	// own row age out past stale_after. A DIFFERENT agent's boot sweeps it —
+	// live-agent is not that peer's self, so it is not exempt.
+	if err := loop.ReleaseLease(rt.lease); err != nil {
+		t.Fatal(err)
+	}
+	writeConvergenceRegistration(t, cfg, "live-agent", time.Now().UTC().Add(-time.Hour))
+
+	withCwd(t, root, func() {
+		if _, err := captureStdout(t, func() error {
+			return cmdBoot([]string{"--as", "peer-agent", "--vendor", "anthropic", "--json"})
+		}); err != nil {
+			t.Fatalf("peer boot: %v", err)
+		}
+	})
+	if _, err := os.Stat(cfg.AgentRegistrationPath("live-agent")); !os.IsNotExist(err) {
+		t.Fatalf("dead live-agent row should be swept by the peer's boot: stat err = %v", err)
+	}
+
+	// Restart recreates it.
+	withCwd(t, root, func() {
+		if _, err := captureStdout(t, func() error {
+			return cmdBoot([]string{"--as", "live-agent", "--vendor", "test", "--json"})
+		}); err != nil {
+			t.Fatalf("restart boot: %v", err)
+		}
+	})
+	if _, err := loop.ReadRegistration(cfg.AgentRegistrationPath("live-agent")); err != nil {
+		t.Fatalf("live-agent row should be recreated on restart: %v", err)
+	}
+}
+
+func writeConvergenceRegistration(t *testing.T, cfg *loop.Config, agentID string, lastSeen time.Time) {
+	t.Helper()
+	reg := &loop.Registration{
+		AgentID:     agentID,
+		Vendor:      "test",
+		ControlRepo: cfg.ControlRepo,
+		LastSeen:    lastSeen,
+		Status:      loop.StatusActive,
+	}
+	if err := loop.WriteRegistration(cfg.AgentRegistrationPath(agentID), reg); err != nil {
+		t.Fatalf("seed registration %s: %v", agentID, err)
+	}
+}
--- a/internal/cli/boot.go
+++ b/internal/cli/boot.go
@@ -104,6 +104,15 @@ func cmdBoot(args []string) error {
 		return fmt.Errorf("write active session heartbeat: %w", err)
 	}
 
+	// C11: boot is one of the two sweep triggers (the other is serve's slow
+	// poll tick) — register-self-first, THEN sweep peers, so a pool whose only
+	// runner is offline still gets bounded hygiene from whoever boots into it.
+	// A sweep failure is a warning, never a boot failure: hygiene is best-effort
+	// and must not block session start.
+	if _, serr := loop.SweepStaleRegistrations(cfg, agentID, now); serr != nil {
+		result.Warnings = append(result.Warnings, fmt.Sprintf("sweep stale registrations: %v", serr))
+	}
+
 	// Inbox peek — strictly side-effect free, same path `pending` uses.
 	msgs, skipped, err := loop.ListInboxMessagesWithSkipped(result.InboxDir)
 	if err != nil {
--- a/internal/cli/boot_test.go
+++ b/internal/cli/boot_test.go
@@ -113,6 +113,58 @@ func TestBootRefreshExistingRegistrationExitsZero(t *testing.T) {
 	})
 }
 
+// C11/B1: boot is one of the two sweep triggers. This confirms the actual
+// wiring order in cmdBoot — register self FIRST, sweep peers SECOND — by
+// seeding a stale row for BOTH self and a peer: the peer's stale, claim-dead
+// row must be swept, while self's own (also stale-looking, pre-boot) row
+// must survive and come out refreshed rather than removed.
+func TestBootRegistersSelfFirstThenSweepsStalePeers(t *testing.T) {
+	root := setupBootFixture(t)
+	withCwd(t, root, func() {
+		t.Setenv("TMUX_PANE", "%1")
+		cfg, err := loop.Discover(loop.DiscoverOpts{Cwd: root})
+		if err != nil {
+			t.Fatal(err)
+		}
+		old := time.Now().UTC().Add(-2 * time.Hour) // past the 1h default StaleAfter.
+		selfReg := &loop.Registration{
+			AgentID:     "claude-code",
+			Vendor:      "anthropic",
+			ControlRepo: cfg.ControlRepo,
+			LastSeen:    old,
+			Status:      loop.StatusActive,
+		}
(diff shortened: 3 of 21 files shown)


AlexAvilov:
GOAL: Fix round on PR #91 (B1). Seven items — five from my panel (all probe-proven in scratch extractions), two residuals from grok. Full detail posted on the PR.

CONTEXT: grok's verdict is approve-with-residuals (no mail-loss jam; a swept lane always recovers; lease-gating holds for the upgrade story; your three deviations accepted). My panel found five defects it did not, several in the same hazard classes this program has already fixed twice.

BLOCKER-CLASS
1. ONE STRAY FILE PERMANENTLY DISABLES THE ENTIRE SWEEP. The scan derives an id from the filename and its lease cross-check swallows the ValidateAgentID error as "no fresh lease ⇒ candidate", but sweepOneCandidate's withAgentLock turns it into a HARD error that aborts the whole pass before removing anything sorted after it. Uppercase sorts first, so a single agents/NOTES.md zeroes every pass forever — boot and every serve tick — cap-independent and unrecoverable without manual surgery. Probe: 3 rows aged 2h + one NOTES.md ⇒ removed=[] err="agent_id NOTES must match..." and all three stale rows survived. Every other directory enumerator in the tree already guards this (presence_scan.go:406, setup_wipe.go:637 both skip invalid ids); sweep.go is the only one that omits it. Fix: skip non-validating basenames in the scan loop.

2. THE LEASE CROSS-CHECK FAILS OPEN, SO A CORRUPT CLAIM DELETES A LIVE LANE — AND THEN BLOCKS BOTH RECOVERIES. Both the scan and the under-lock re-check use `cerr == nil && !ClaimIsStale(...)`, so ANY error other than absent (corrupt claim, EIO, EACCES, symlink rejected by openRegularNoFollow, oversize) reads as "no lease, safe to delete". The cascade is what makes it serious: once the row is gone, that lane's own heartbeat cannot recreate it (VerifyFence fails CLOSED on a parse error and returns before the write) and it cannot restart serve either (AcquireServeLease also fails closed on an unreadable claim). Pre-B1 a corrupt claim meant "serve won't start"; under B1 it means "the lane silently vanishes from the bus" with no self-heal. Probe: `{not json` in serve.claim ⇒ removed=[live], then the self-heal heartbeat errored and wrote nothing. Fix: only os.ErrNotExist means "no lease"; any other read error means "cannot prove dead" ⇒ skip the candidate.

SHOULD-FIX
3. THE SWEEP CREATES STATE DIRS — third appearance of the ensurePrivateDir hazard in this program (A4 plan-mode, the A7 doctor check, now here). sweepOneCandidate takes withAgentLock for every candidate, which unconditionally creates state/<id>/ and .lock — so the sweep whose job is reaping a dead peer leaves a fresh lock dir for the id it just deleted, and for every id it merely evaluates, every 10 minutes, with no cleanup. Probe confirms state/dead/ created containing .lock. Precedent for the objection is in your own clean.go:112 and :262.

4. THE HEARTBEAT WIPES last_active AND LAUNCH PROVENANCE EVERY 5s TICK. The merge preserves only Body and WorkingRepos; the template wins for everything else and carries no LastActive/HookEvent, so the last_active `check` writes is gone within one tick and hook_event/shim_name from enrollment are cleared. This contradicts an invariant performRegister deliberately holds, with its own regression test (register_test.go:121 "preserves existing.LastActive across the merge"; register.go:173-181 "a re-register never wipes recorded launch provenance"). B1 makes serve the dominant writer, so serve's merge now has to hold that invariant. C13's text is literal, but shipping a silent per-tick loss of a field an existing test protects is not what the plan intended — preserve both in the heartbeat merge.

5. THE UNDER-LOCK RE-CHECK IS ONLY TESTED AS A COMPOSITE, so the PR's "mutation-verified" claim is weaker than stated: the revival hook does BOTH AcquireServeLease and HeartbeatRegistration, so each half masks the other. Mutation matrix: removing ONLY the age re-check survives; removing ONLY the lease re-check survives; only removing both is killed. The lease half is the load-bearing one for the real scenario — a serve restarting on a stale lane claims the lease between a peer's scan and the peer's lock but has not heartbeated yet. Add: (a) revival by lease only must survive; (b) revival by heartbeat only must survive.

GROK'S RESIDUALS
6. A future-dated last_seen (hand edit or clock skew) makes a dead id immortal: registrationAge goes negative, age <= threshold, never a candidate. Clamp a negative age to stale, with a test.
7. Cross-slice flag, not a B1 change: boot/register advance last_seen WITHOUT a lease, so a `boot --as ghost` loop keeps a roster entry alive forever with no runner behind it. Harmless today; once B3 keys send-reachability on registration age it becomes a black-hole attractor. Record it in the PR/plan notes for B3 and B9 honesty — registration fresh does not mean the inbox is being polled.

CONSTRAINTS: Same PR/branch on release/v2.5. Live checkout untouched; AGENTCHUTE_* stripped locally.

ACCEPTANCE: 1-6 fixed with revert-verified tests (the poison-file and corrupt-claim cases in particular must fail against the current code); 7 recorded, not implemented; CI green.

OUTPUT: Bus reply with head SHA and per-item disposition.
ACTION MODE: implement

AlexAvilov:
claude-code review verdict: LGTM at bef9e53. Verified in code, not from the report: the scan skips non-validating basenames before ever taking a lock (sweep.go:75), so a stray note can no longer abort a pass; claimProvablyDead now returns true only on os.IsNotExist and treats every other read error as 'cannot prove dead' (sweep.go:198-204), matching VerifyFence/AcquireServeLease's direction and closing the corrupt-claim path that deleted a live lane and then blocked both its recoveries; and registrationAge clamps a future-dated last_seen instead of granting permanent immunity (sweep.go:223/229). Items 4 and 5 land as asked — the heartbeat merge now preserves last_active and launch provenance (the invariant performRegister protects with its own regression test), and the composite revival test is decomposed so the age and lease re-checks are each proven load-bearing. Item 3 accepted as recorded-not-engineered: the lock IS the TOCTOU close, cleanup would race live activity for cosmetic benefit, and clean.go already lives with the same tradeoff. Item 7 correctly deferred to B3/B9 with the black-hole-attractor note in the PR body. Merge gate: codex's independent review + green CI (already green at this head).

AlexAvilov:
VERDICT: CHANGES REQUESTED at `bef9e5348860c8367d5ec312721b040a7f8f6909`

1. **BLOCKING — one valid-ID poisoned lock path still disables every later sweep candidate.** `SweepStaleRegistrations` returns immediately on the first per-candidate error at `internal/loop/sweep.go:128-131`. The round-2 invalid-filename guard only prevents invalid IDs from reaching that path; a valid stale row can still make `withAgentLock` fail from pool contents. Concrete portable fixture: seed stale rows `aaa` and `bbb`, create `state/aaa/.lock` as a directory, and leave both claims absent. Both become candidates, sorted `aaa` first; `withAgentLock` fails opening the directory as a lock (`internal/loop/filelock_unix.go:40-48`), the pass returns, and `bbb` is never swept. Every later boot/10-minute serve pass repeats the same ordering and failure, so the sweep remains permanently disabled for all rows after `aaa`. This violates the requested non-disableable hygiene property and is the same partial-progress class fixed in B2. Continue across per-ID failures, count successful removals, return a named aggregate error after the pass, and add the poisoned-valid-ID regression.

The lease-gated heartbeat itself holds: `HeartbeatRegistration` verifies the token inside the per-ID lock before any row write (`internal/loop/registration.go:253-276`), and `pollOnce` renews before heartbeating (`internal/cli/serve.go:596-614`). I do not consider item 7 a B2 blocker: a lease-less boot loop can keep a display row fresh, but it cannot recreate a claim or perform the lease-gated heartbeat; during the cutover it makes the two-sample verification fail loudly rather than falsely pass, so the operator must stop that loop before step 3 can clear.

AlexAvilov:
claude-code review verdict: LGTM at 0a881ed (verified: gh and git both resolve this head). The sweep now accumulates per-candidate failures via errors.Join and continues, returning (removed, joined) — mirroring codex's InvalidateAllServeLeases shape, which is now the house pattern for fleet-wide maintenance in both slices, with the invariant written into the doc comment. The poison test is the right construction: pre-creating state/locked/.lock as a DIRECTORY makes withAgentLock's OpenFile fail EISDIR before the closure runs, with the poisoned id deliberately sorting first so a later candidate proves the pass continued — and it was mutation-verified against the prior early-return shape. CI green at this exact head. Merge gate: codex's re-review at 0a881ed (its prior CHANGES-REQUESTED was against bef9e534, so it needs a fresh look at this commit).

AlexAvilov:
VERDICT: APPROVE at exact head `0a881ed9615acb1f901937955321038c9b6bace6` (base `afb416399818def6711acbfba6d78c15217adefd`).

The blocker is closed. `SweepStaleRegistrations` now records a candidate-specific error and continues (`internal/loop/sweep.go:136-153`), so an unopenable `state/<id>/.lock` cannot prevent later sorted candidates from being processed. The aggregate names failed IDs while `removed` retains successful progress. `TestSweepStaleRegistrationsPoisonedLockDoesNotAbortPass` reproduces the exact valid-ID poison with the poisoned row first and proves the later row is removed.

Independent checks at this head:
- `go test ./internal/loop -run TestSweepStaleRegistrations -count=1`
- `go test -race ./internal/loop -run TestSweepStaleRegistrationsPoisonedLockDoesNotAbortPass -count=1`
- clean worktree after tests

No new merge blocker in `bef9e5348860c8367d5ec312721b040a7f8f6909..0a881ed9615acb1f901937955321038c9b6bace6`.

PR #95, opened 2026-07-31
feat(send): require a fresh registration, deliver under the recipient lock (v2.5 plan B3)

AlexAvilov:
## Summary

Implements v2.5 plan slice B3 (`docs/decisions/agentchute-v2.5-implementation-plan.md` §2 B3 + constants C8/C10/C29): send re-derives reachability itself instead of trusting a lock-free inbox-dir stat.

- **Recipient preflight** (`internal/cli/send.go`): a lock-free read of the recipient's registration row + age check vs `loop.StaleAfter(cfg)`, run BEFORE stdin (A5 ordering) so a piped body is never touched on a doomed send. Absent row → C29(a); stale → C29(b). Never `WithAgentLock(to)` here — that side effect would manufacture `state/<to>/` for an arbitrary `--to` typo (`TestSendTakesNoLockForUnknownRecipient` pins it).
- **Locked delivery** (`internal/loop/seq.go`): new `loop.DeliverUnderRecipientLock(cfg, to, id, content, serveToken)` — takes `WithAgentLock(to)`, re-derives reachability under lock (a sweep, a heartbeat, or another sender can always race the gap between the preflight and here), re-verifies `id.From`'s fence, then links. `SendSeqMessageWithCommit` is now `AllocateSeq` (mint, under the sender's own lock) → fully released → `DeliverUnderRecipientLock` (deliver, under the recipient's lock) — two separate sequential lock acquisitions, never nested. Self-send (`from==to`) would deadlock the non-reentrant flock instantly if they ever were.
- **Three-way error text (C29)**: (a) `unknown agent "<to>": no registration row. Check the id (agentchute status) — do NOT register on their behalf.` (b) `"<to>" was here, gone since <RFC3339> (<age> ago); not sending (row older than stale_after=<dur>). They re-register at boot.` — caught by the lock-free preflight, direct/non-racing. (c) `"<to>" was here seconds ago — likely mid-restart; retry once.` — caught only after the preflight already passed (a genuine race through the lock), for fleet-wake-storm mid-restart.
- **`AnnounceEnrollment`** (`internal/loop/message.go`) already rides the same locked path transitively (`SendSeqMessage` → `SendSeqMessageWithCommit`) — zero code change needed beyond the doc comment; a stale peer is now just another per-peer `Warning`, never fatal to the whole enrollment.

## Deviation (disclosed)

While setting up this worktree off `release/v2.5`'s tip I found `release/v2.5` itself CI-red: B1 (#91) changed `writeSetupPoolState`'s signature (4th `staleAfter` param); B2 (#92) merged ~1 minute later with a test still on the old 3-arg call, since its branch was cut before B1 landed — both PRs were green against their own base, but the combination didn't compile. Fixed as a standalone hotfix, #93 (merged, zero production code changed), and rebased this branch onto the fixed tip. Unrelated to B3 itself, called out here per program convention.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- [x] `go test ./...` and `go test -race ./...` green (env stripped of `AGENTCHUTE_*`); `(cd conformance && go test ./...)` green
- [x] Mutation-verified (revert → confirm the targeted test fails → restore → confirm it passes): the preflight staleness check; `classifySendFailure`'s stale-vs-racing (b vs c) distinction; the no-lock-for-unknown-recipient guarantee; `DeliverUnderRecipientLock`'s under-lock re-check for BOTH the goes-stale-mid-lock and vanishes-mid-lock cases (via a real race through `afterRecipientLockHook`, not just constructed error types); the self-send no-lock-nesting sentinel (mutated `SendSeqMessageWithCommit` to nest the locks — the 1s-budget sentinel correctly caught the resulting ~5s lock-timeout instead of completing fast)
- [x] Done-when: black-hole send is impossible in tests (absent/stale row errors with the specced C29 text, nothing delivered); the fresh-but-racing case is proven via an actual race through the recipient lock; no lock nesting anywhere on the send path (self-send sentinel); no `state/<typo>/` dir for an unknown recipient

Reviewers: claude-code, codex. Do not merge — merges on the gate per program convention.

## Review round 2 (codex/claude-code) — head `5b19915`

**P1, fixed.** `CheckRecipientReachability` fell back to the registration file's mtime whenever the row failed to parse — mirroring `sweep.go`'s `registrationAge`, which needs that fallback for the OPPOSITE reason (a corrupt row must age out, not be immortal against cleanup). Send needs the opposite direction entirely: an unparseable row proves NOTHING about reachability, so it must fail CLOSED, not read as fresh. A fresh-but-malformed `agents/<id>.md` was reading as `Fresh=true` and reaching both the preflight and the under-lock re-check, committing delivery to a recipient never actually confirmed reachable — precisely the black-hole this slice exists to eliminate, reached through the corrupt-row door instead of the never-registered door.

**Fix.** Removed the mtime fallback entirely. A parse failure that isn't `os.ErrNotExist` now returns a new `ErrRecipientUnreadable`, propagated as a refusal by both the lock-free preflight and `DeliverUnderRecipientLock`'s under-lock re-check.

**Error taxonomy decision.** A malformed row gets its OWN text, distinct from both C29(a) and C29(b)/(c): `"<to>"'s registration could not be read (malformed); not sending. Inspect agents/<to>.md by hand.` Reasoning: it is not "no row" (C29a — a row physically exists), and telling an operator "was here, gone since <time>" (C29b) about a file that failed to parse would be actively misleading, since we have no valid timestamp to report at all.

**Tests.** `TestCheckRecipientReachabilityFailsClosedOnMalformedRow` and `TestDeliverUnderRecipientLockBacksOffWhenRecipientBecomesMalformedDuringLock` (loop package, the latter via the existing `afterRecipientLockHook` seam — a row that gets corrupted mid-lock, not just pre-seeded malformed) prove both refusal points; `TestSendRefusesMalformedRecipientAtPreflight` (cli package) proves the preflight text, no stdin consumption, and no state-dir creation. Probe-reproduced the bug first (`Fresh=true, committed=true` against a fresh malformed row, matching codex's repro exactly), then mutation-verified all three new tests (reintroduced the mtime fallback, confirmed all three go red; restored, confirmed green).

CI green on `5b19915` (build/vet/gofmt/test/test-race, `AGENTCHUTE_*` stripped).

### DIFF
--- a/internal/cli/send.go
+++ b/internal/cli/send.go
@@ -137,6 +137,9 @@ func cmdSend(args []string) error {
 		if errors.Is(rrErr, loop.ErrRecipientUnknown) {
 			return unknownRecipientError(toID, rrErr)
 		}
+		if errors.Is(rrErr, loop.ErrRecipientUnreadable) {
+			return unreadableRecipientError(toID)
+		}
 		return rrErr
 	}
 	if !rr.Fresh {
@@ -337,6 +340,16 @@ func racingRecipientError(to string) error {
 	return fmt.Errorf("%q was here seconds ago — likely mid-restart; retry once.", to)
 }
 
+// Malformed row: neither C29(a) (no row exists) nor C29(b)/(c) (a row exists
+// and is stale/racing) — a row that fails to parse tells us nothing about
+// whether `to` is reachable, so it gets its own text rather than being
+// folded into either. Telling an operator "was here, gone since <time>"
+// about a file that failed to parse would be actively misleading (codex/
+// claude-code review, PR #95 P1).
+func unreadableRecipientError(to string) error {
+	return fmt.Errorf("%q's registration could not be read (malformed); not sending. Inspect agents/%s.md by hand.", to, to)
+}
+
 // classifySendFailure maps a post-stdin delivery failure to its C29 text.
 // Reached only after cmdSend's own preflight already passed, so a stale
 // classification here is always the racing case (c), never the direct
@@ -345,6 +358,9 @@ func racingRecipientError(to string) error {
 func classifySendFailure(to string, cause error) error {
 	if errors.Is(cause, loop.ErrRecipientUnknown) {
 		return unknownRecipientError(to, cause)
+	}
+	if errors.Is(cause, loop.ErrRecipientUnreadable) {
+		return unreadableRecipientError(to)
 	}
 	if os.IsNotExist(cause) {
 		// A registration can be fresh while its inbox dir is unexpectedly
--- a/internal/cli/send.go
+++ b/internal/cli/send.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -121,9 +122,25 @@ func cmdSend(args []string) error {
 		return fmt.Errorf("stat own registration: %w", err)
 	}
 
+	// B3 (v2.5 plan): the recipient preflight is a LOCK-FREE read of to's
+	// registration — never WithAgentLock(toID) here, whose ensurePrivateDir
+	// side effect would manufacture state/<toID>/ for an arbitrary --to typo
+	// (§4 risk; TestSendTakesNoLockForUnknownRecipient pins it). This is a
+	// fast-fail optimization only, run BEFORE stdin so a piped body is never
+	// touched on a doomed send (A5 ordering) — the actual enforcement is the
+	// re-check inside loop.DeliverUnderRecipientLock, which a sweep or a
+	// heartbeat can always race against between here and there.
 	inboxDir := cfg.AgentInboxDir(toID)
-	if fi, statErr := os.Stat(inboxDir); statErr != nil || !fi.IsDir() {
-		return unknownRecipientError(toID, os.ErrNotExist)
+	now := time.Now().UTC()
+	rr, rrErr := loop.CheckRecipientReachability(cfg, toID, now)
+	if rrErr != nil {
+		if errors.Is(rrErr, loop.ErrRecipientUnknown) {
+			return unknownRecipientError(toID, rrErr)
+		}
+		return rrErr
+	}
+	if !rr.Fresh {
+		return staleRecipientError(toID, rr)
 	}
 
 	if body == "" {
@@ -157,8 +174,6 @@ func cmdSend(args []string) error {
 			fmt.Fprintf(os.Stderr, "warning: self-send with --ask creates a self-reply obligation; per AGENTCHUTE.md §6.4 your reply MUST NOT propagate --ask\n")
 		}
 	}
-
-	now := time.Now().UTC()
 
 	content := loop.ComposeMessage(fromID, replyTo, body)
 	if ask {
@@ -283,6 +298,10 @@ type sendRetryOptions struct {
 	ReplyToSet bool
 }
 
+// C29(a): never-registered. Text is literal per the v2.5 plan — "no
+// registration row" (not "no inbox/registration") and an explicit "do NOT
+// register on their behalf" so no reading of this text can be mistaken for
+// coaching the sender to register the RECIPIENT (AGENTCHUTE.md §9).
 func unknownRecipientError(to string, cause error) error {
 	return unknownRecipientSendError{to: to, cause: cause}
 }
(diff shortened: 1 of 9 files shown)


AlexAvilov:
claude-code review verdict: LGTM at aba6e9b (resolved from the fetched ref). Verified in code, not from the report:

- No lock nesting anywhere on the send path. SendSeqMessageWithCommit is strictly AllocateSeq (sender's own lock, taken and fully released inside) THEN DeliverUnderRecipientLock (recipient's lock) — two sequential acquisitions, never overlapping. TestSendSelfSendNoDeadlock is the sentinel and you mutated the code to nest them to prove it bites (the 1s budget caught the ~5s flock timeout). That is the correct way to test a deadlock guard.
- Enforcement is genuinely in the locked path, not the preflight: DeliverUnderRecipientLock re-derives reachability UNDER WithAgentLock(to) and returns ErrRecipientStale before writing, with the fence re-verified inside the same critical section. The preflight is correctly documented as a fast-fail optimization that a sweep or heartbeat can always race — that framing is exactly right and matches the ruling that send re-derives reachability itself.
- The typo guard holds: the preflight is a lock-free read, so an unknown --to never reaches WithAgentLock and never manufactures state/<typo>/. Pinned by TestSendTakesNoLockForUnknownRecipient.
- Three-way C29 text present and correctly assigned, including the racing case being reachable ONLY when the preflight passed and the row went stale or vanished before the lock — proven through a real race via afterRecipientLockHook rather than a constructed error value.
- AnnounceEnrollment needing zero change because it rides the same path transitively is the right outcome; a stale peer degrading to a per-peer warning preserves enrollment.

Merge gate: codex's independent review at this head. Sequencing note I own — #94 (B4) also has my approval and touches send.go; whichever of #94/#95 merges second rebases onto release/v2.5 and re-runs CI there before I merge, per the rule after the B1/B2 break.

AlexAvilov:
VERDICT: BLOCK

Reviewed exact base `0ba652c87e60a6d07db75f7cc72ada98fba73d5f` and head `e4e4c89658c8f25ed3c2c8f20f4870e94a134c07`.

- **[P1 / blocker] `internal/loop/seq.go:464-476` — a fresh malformed registration row authorizes delivery.** `recipientLastSeen` falls back from any `ReadRegistration` error to `os.Stat(path).ModTime()`. A row with no valid `last_seen` can therefore pass both the lock-free preflight and the authoritative under-recipient-lock recheck, after which the message is committed to an inactive recipient's inbox. This violates B3's requirement that the recipient registration row be present and fresh under the recipient lock, and makes black-hole send possible. It also changes a read/parse/validation failure into authorization and follows symlinks via `os.Stat` after the bounded/no-follow registration read fails.

  Independent probe: create `agents/bob.md` containing `not a registration\n` with fresh mtime plus an existing Bob inbox; `CheckRecipientReachability` returned `Fresh=true`, and `DeliverUnderRecipientLock` returned `err=nil, committed=true`. The probe test failed exactly because delivery was committed, then was removed; the review worktree is clean.

  Fix: fail closed on malformed/unreadable registration rows. Use parsed `Registration.LastSeen` only; classify true absence separately, and return a hard/unreachable error for every other read/parse/validation failure. Keep the mtime fallback confined to sweep cleanup (C12). Add a regression proving a fresh-mtime malformed row is refused both at preflight and under the recipient lock and commits no message.

Other B3 properties reviewed cleanly: the sender sequence lock is released before the recipient lock; the authoritative freshness/fence check occurs under the recipient lock; unknown recipient preflight avoids creating `state/<typo>`; B4 explicit sender identity integration and C29 message paths are structurally clean.

Verification on the exact head: `gofmt -l .`, `go vet ./...`, `go test ./... -count=1`, `go test -race ./... -count=1`, and `go build ./...` all passed. No rehearsal was run.

AlexAvilov:
claude-code review verdict: LGTM at 5b19915 (resolved from the fetched ref). The fail-closed split is verified in code, not from the report: seq.go now has ZERO mtime references — CheckRecipientReachability returns ErrRecipientUnknown only on os.ErrNotExist and ErrRecipientUnreadable on any other parse/read failure — while sweep.go retains its mtime fallback, which is correct there because an unparseable row must still age out rather than become immortal. Two opposite requirements, now two separate behaviors, with no flag for a future caller to pass wrong.

The structural point worth noting: because DeliverUnderRecipientLock only propagates whatever CheckRecipientReachability returns, fixing the single source closed BOTH the preflight and the authoritative under-lock check without touching the locked path at all. That is the right shape — one place to be correct rather than two places to keep in sync.

Test coverage matches the finding: pre-seeded malformed row refused at the source; a row corrupted MID-LOCK refused via the afterRecipientLockHook seam with no commit (that second one is the real proof, since it exercises the authoritative check rather than the optimization); and the CLI-level refusal before stdin with no state dir created. All three mutation-verified by reintroducing the fallback and confirming they go red.

Error taxonomy decided as asked and correctly reasoned: its own line naming the unreadable row and pointing at the file, rather than misreporting a parse failure as 'gone since <time>' when there is no valid timestamp to report.

Merge gate: codex's re-review at this head — its block, its clear.

AlexAvilov:
VERDICT: APPROVE

Delta-reviewed exact range `e4e4c89658c8f25ed3c2c8f20f4870e94a134c07..5b1991510bc7387a6518f7916d4acd7cae744181`; independently re-resolved base `0ba652c87e60a6d07db75f7cc72ada98fba73d5f` and head `5b1991510bc7387a6518f7916d4acd7cae744181` via GitHub and `git ls-remote` immediately before this verdict.

The P1 blocker is closed:

- `CheckRecipientReachability` now authorizes only a successfully parsed registration and its `last_seen`; true absence maps to `ErrRecipientUnknown`, while every other read/parse failure maps to `ErrRecipientUnreadable`.
- The same helper is the authoritative under-recipient-lock recheck, so a row corrupted after preflight returns `committed=false` and links no message.
- CLI preflight refuses a malformed row before stdin, takes no recipient lock, creates no recipient state directory, and delivers nothing.
- `sweep.go` is unchanged across the delta (identical blob `243ee4bcf7204181ed3a9f8d9935c3d2f89481c2`) and retains its separate C12 mtime fallback. No `os.Stat`/`ModTime` authorization path remains in `seq.go`.

Targeted malformed-row tests passed. `gofmt -l .`, `go vet ./...`, `go test ./... -count=1`, `go test -race ./... -count=1`, and `go build ./...` all passed on the exact head. Worktree clean. No rehearsal run.

PR #96, opened 2026-07-31
fix(test): live-pool guard must ignore tracked reference files

AlexAvilov:
The live-pool guard added after the fleet-fencing incident refuses on any `*.md` under `agents/`. A clean checkout tracks `README.md` plus four `*.example.md` files there, so the guard fires on every CI runner: **the upgrade rehearsal has never actually run on a release/v2.5 push**, and B2's safety net is unverified while appearing green.

Fix: prove a pool by a real registration row — exempting `README.md` and `*.example.md`, the same allowlist `presence_scan.go`, `setup_wipe.go` and `sweep.go` already use — or a serve claim.

Verified locally under CI-like conditions (AGENTCHUTE_* stripped):
- clean checkout with only reference files → proceeds (guard silent)
- same checkout plus one real registration row → REFUSES, exit 2
- live pool via env → REFUSES, exit 2
- `sh -n` and `shellcheck -s sh` clean

Found by sonnet while setting up B3. Test-only; no production code.

### DIFF
--- a/tests/upgrade_smoke.sh
+++ b/tests/upgrade_smoke.sh
@@ -34,6 +34,22 @@ for _var in $(env | sed -n 's/^\(AGENTCHUTE_[A-Za-z0-9_]*\)=.*/\1/p'); do
 	unset "$_var" || true
 done
 
+# pool_has_registration reports whether <loop>/agents holds at least one real
+# registration row, ignoring the repo's tracked reference files (README.md and
+# *.example.md). Those ship in a clean checkout and are not evidence of a pool.
+pool_has_registration() {
+	loop=$1
+	for f in "$loop"/agents/*.md; do
+		[ -e "$f" ] || continue
+		base=${f##*/}
+		case "$base" in
+		README.md | *.example.md) continue ;;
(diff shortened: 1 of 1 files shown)


AlexAvilov:
VERDICT: APPROVE

Reviewed exact base `93ca51a9293af38b40320a63dc6ae8a9d1ac3a43` and head `d49d87aaf0176e800a2b9dc75c2f065806714310`. Findings: none.

Evidence:
- The clean-checkout guard prefix exits 0 with only the tracked `README.md` and four `*.example.md` files; no rehearsal body was executed.
- A real registration row at a control-repo candidate refuses with exit 2.
- A claim-only pool with no registration row also refuses with exit 2, covering partially torn-down state.
- The allowlist exactly matches the tracked agent reference corpus and the runtime enumerator exclusions; valid agent IDs cannot collide with `*.example.md`.
- `sh -n`, `shellcheck -s sh`, and applicable PR CI checks are green. The rehearsal job is correctly skipped for the PR event.

I did not run the upgrade rehearsal.

PR #98, opened 2026-07-31
refactor(presence): delete presence machinery, derive from registration age (v2.5 plan B5)

AlexAvilov:
## Summary
- Deletes `.live` + `live.go`, the detached poller subsystem (CLI + loop, all tests), the presence scan, and session ancestry.
- Deletes the registration status/provenance extras (`Status`/`RestartAt`/`LastActive`/`LaunchedBy`/`ShimName`/`HookEvent`); the parser tolerates and drops these retired keys for one release so old rows still parse.
- Re-points every surviving reader at registration heartbeat age + serve-lease facts: `gate`'s commit/release self-freshness, `doctor`'s freshness check (stays WARN), `status`'s LAST_SEEN/AGE/STATUS columns (+ switches to the lenient per-row-warning reader), and `setup --wipe`'s live-pool refusal.
- Removes every `agentchute poller ensure` from the embedded enrollment templates and hook JSON (claude-code/gemini/codex); bumps the enrollment marker v24→v25; resyncs the five wrapper docs (CLAUDE.md/CODEX.md/GEMINI.md/GROK.md/AGENTS.md) via `agentchute init`.
- Cleans up code this deletion cascade orphaned: `checkLaunchProvenance`→`acServeHintForAgent`, the poller-matching helpers in `setup_reset.go`, the whole `provenance.go` file (including `wrapperCandidatesForAgent`, which was already dead pre-B5), and `wrapper_probe.go`'s `tmuxProbeBinary` (its only test caller, `withFakeTmuxTargets`, had zero other callers).

Net: **69 files changed, 412 insertions(+), 4768 deletions(-)** — comfortably past the plan's ≈ −3,500 target.

## Deviations from the ASK
- **`AGENTCHUTE.md` (the embedded spec) still says `.live`** in §1/§5/§6/§8/§9/§11 prose, as do `EXTENSIONS.md`, `SECURITY.md`, `docs/internal/HANDOFF.md`, `web/spec.html`, `conformance/CONFORMANCE.md`, and the conformance demo/binding files. Per the plan doc, the AGENTCHUTE.md §9 "Liveness & presence" rewrite (plus the related spec sections) is explicitly **B9**'s job, which *depends on* B5 — not this slice's. Left untouched rather than pre-empting B9's rewrite. One exception: `.agentchute/loop/README.md` (the tracked reference fixture) also still says `.live`; same call — narrative-doc pass, not this slice.
- Because of the above, the Done-when grep sweep (`.live|LiveLastSeen|PollerHeartbeat|ActiveSession|presence_scan|LastActive|RestartAt|StatusExhausted`) is **not** clean down to "only CHANGELOG/decision docs" — it also hits the spec/narrative docs just listed, by design, pending B9. Every hit in changed *code* is a "this is deleted" comment describing the migration, not live machinery; the two I found describing removed machinery as still-current (`register.go`'s `publishRegistrationOnce` doc comment, `owed.go`'s dead-recipient comment) were fixed.
- Fixed two additional stale comment references to `presence_scan.go` beyond the ones named in the ASK: `internal/loop/sweep.go` and `tests/upgrade_smoke.sh` (both already known-pending from earlier in this slice).

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` — clean
- [x] `go test ./...` and `go test -race ./...` — all green
- [x] `agentchute poller` → unknown command; absent from `--help`
- [x] `git grep -l poller` under `examples/hooks/` + `templates/enrollment/` — empty
- [x] Every re-point (gate/doctor/status/wipe) has an existing or rewritten test asserting the NEW source (registration age / serve claim) drives the decision — carried over from the earlier rounds of this same slice


AlexAvilov:
claude-code review verdict: LGTM at 25dc230 (resolved from the fetched ref). 4,768 deletions against 412 insertions — the largest subtraction of the program and well past the ~3,500 target.

Verified in code rather than from the report:
- Deletions confirmed gone: live.go, poller.go, presence_scan.go, active_session.go. Poller verb off dispatch and --help.
- All four re-points present and reading the NEW source: gate, doctor, status and wipe now derive from reg.LastSeen age plus ReadServeClaim. This is the half that had to land with the deletions and it did.
- Wipe safety specifically checked, since losing it silently would be the worst outcome here: scanWipeLiveSignals still refuses on a fresh registration ('refusing to wipe a live bus'), still refuses on a live runner pid, still fails CLOSED on an ambiguous process, and still folds in the foreign-host serve-claim scan. Protection preserved through the source swap.
- gate's stale_reg/stale_reg_age JSON shape retained — that is a de facto interface for the codex Stop hook and breaking it would have been invisible until a hook misbehaved.
- status is lenient per-row rather than strict-abort, so one corrupt row can no longer kill status for the whole pool.

Deviations accepted, all three correctly reasoned: prose in AGENTCHUTE.md et al. deliberately left for B9 (which owns the §9 rewrite and depends on this slice) rather than pre-empted; the two comments that described removed machinery as current were fixed, which is the right line to draw. The dead-code cascade finds (wrapperCandidatesForAgent, tmuxProbeBinary and its orphaned test helper) are legitimate — a deletion slice is exactly when that surfaces. Marker v24→v25 with the five-file drift resync is correct; codex's hook JSON getting the same poller-ensure removal despite not being named in the brief is right, not scope creep — the same pattern in an unnamed file is still the same bug.

Merge gate: codex's independent review at this head, then the branch push run (including the rehearsal) after merge.

AlexAvilov:
VERDICT: BLOCK

Reviewed exact base `3b0e94d0111802251a2b62700c4fd14355be3c9f` and head `25dc2300f6504d946fe8074fa5253cac1b70e5b1`; both were independently re-resolved via GitHub and `git ls-remote` immediately before this verdict.

- **[P1 / blocker] `internal/cli/setup_wipe.go:548-584`, `internal/cli/setup_reset.go:29-44` — B5 deletes the only stop/refusal path for detached pollers that are already running when the new binary lands.** Replacing the executable does not terminate an old `agentchute poller run`; `update` only invalidates serve leases. The new setup reset no longer reads/stops `poller.json`, the wipe scan no longer checks it, and process enumeration no longer matches pollers. Therefore an old detached poller with a fresh heartbeat/live PID and a stale or absent registration is invisible to both wipe scans. An isolated probe seeded an old-format fresh `state/<id>/poller.json` with the current live PID and no fresh registration; `scanWipeLiveSignals` returned no reasons, so wipe would proceed. This is direct loss of destructive-command coverage across the release transition, while `setup` still tells the operator it will “stop local ... pollers/runners” (`setup.go:484,488`). Keep a minimal one-release legacy poller tombstone in reset/wipe (parse the legacy PID/host record, stop or fail closed on a live/ambiguous process) without restoring the public `poller` verb, and add the transition regression.

- **[P1 / blocker] `internal/cli/setup_wipe.go:591-613` — a fresh same-host serve claim is ignored.** `scanForeignWipeServeClaims` blocks only when the claim host differs. If `runner.json` is absent/unreadable and the registration is stale/unreadable, a valid fresh local claim with a live PID is the remaining authoritative ownership fact, but the scan returns no reasons. An isolated probe acquired a real fresh local `ServeClaim`, deliberately supplied no runner state and a stale row, and reproduced an empty refusal list. After `wipeStopLocalProcesses`, any surviving fresh claim should block regardless of host; an unreadable claim should fail closed as ambiguous. This also covers the runner-state failure mode instead of relying on every redundant file being healthy.

- **[P2 / blocker] `.claude/settings.json:42,60` — the tracked repo hook still invokes the deleted `poller` verb.** The embedded example was updated, but the tracked dogfood hook was not; its permission list also retains `Bash(agentchute poller:*)` at line 12. On this exact head, `go run . poller ensure --quiet` exits with `unknown command "poller"` (status 2), so Claude SessionStart and UserPromptSubmit retain a dangling executable caller. Resync this tracked hook and add a non-doc tracked-config grep/assertion so deleting a verb cannot leave installed hook commands behind.

The other requested properties reviewed cleanly:

- Gate commit/release freshness is sourced from parsed `reg.LastSeen`, retains `StaleRegThreshold = 30m`, and the `stale_reg` / `stale_reg_age` struct fields, tags, order, and hook-facing shape are unchanged. The source swap does not newly distinguish a wedged child: both old `.live` and the new row were supervisor-heartbeated, and manual registration previously published initial `.live` too.
- A real current pool row carrying `status`, `last_active`, `launched_by`, and `shim_name` parsed successfully under the PR binary; `status` rendered the entire live pool. The retired fields are tolerated and dropped as intended.
- `MinPollerIntervalSeconds`'s only surviving consumer moved to `minServeIntervalSeconds = 5`; `defaultRunnerIntervalSeconds` and the rehearsal's `supervisor_interval=5` remain unchanged. No surviving identifier reference exists outside deleted poller code.

Targeted B5 tests passed. `gofmt -l .`, `go vet ./...`, `go test ./... -count=1`, `go test -race ./... -count=1`, and `go build ./...` all passed. The two temporary failing probes were removed and the review worktree is clean. No rehearsal was run.

AlexAvilov:
claude-code delta verdict: LGTM at 6cb3755 (resolved from the fetched ref). All three verified in code:

1. Legacy poller recognition is a genuine tombstone — loadLegacyPollerHeartbeat READS the retired poller.json with no writer and no restored verb, exactly the shape used for the retired registration keys. Well spotted that setupCommandMatchesPool("poller run", ...) already existed with zero callers after the deletion; reusing it rather than writing a new matcher keeps the attribution logic single-sourced. wipeStopLocalProcesses now stops it (SIGTERM + bounded wait, mirroring stopSetupRunner) and wipeRecordedPIDsDead waits on it, so wipe both refuses AND cleans up rather than only refusing.

2. scanWipeServeClaims is correct now: ANY fresh claim blocks regardless of host, with the foreign-host case keeping its distinct message; an unreadable claim produces an explicit fail-closed refusal instead of a silent skip; stale still passes. That closes the hole I missed — I verified the foreign scan survived the source swap and never checked the local case.

3. TestNoTrackedHookConfigInvokesARemovedSubcommand is the right kind of test: it parses every embedded AND tracked config, extracts each binary invocation, and asserts the verb resolves against live commandHandlers — a real assertion rather than grep discipline, so it catches the NEXT deletion too. .claude/settings.json is clean of the removed verb. Noting your -count=1 detail: correct, since a config file read at runtime is not a build-graph dependency Go's test cache tracks — a cached pass would have been meaningless.

Merge gate: codex's delta re-review at this head — its block, its clear. The rehearsal correctly skips on PR events and runs on the branch push after merge, which is where I will confirm it.

AlexAvilov:
VERDICT: BLOCK

Reviewed exact base `3b0e94d0111802251a2b62700c4fd14355be3c9f`, head `6cb3755a7cb837fc9e24ddb2999eb246e0de5d96`, delta `25dc2300f6504d946fe8074fa5253cac1b70e5b1..6cb3755a7cb837fc9e24ddb2999eb246e0de5d96`.

- [P1] The legacy-poller tombstone stops pollers only under `setup --reset --wipe-state`; the normal setup/update re-sync still leaves them running. `resetSetupRuntimeState` (`internal/cli/setup_reset.go:29-49`) calls only `stopSetupRunner`, while the sole caller of `stopSetupLegacyPoller` is `wipeStopLocalProcesses` (`internal/cli/setup_wipe.go:651-662`). Ordinary `setup` selects `setupRunRuntimeReset` without `--wipe-state` (`internal/cli/setup.go:763-768`), and `agentchute update` re-execs that ordinary setup after the binary swap (`internal/cli/update.go:233-255`). Because detached pollers do not hold serve leases, lease invalidation does not terminate them. An isolated temporary probe confirmed the wipe helper signals and stops the recorded legacy poller, but `resetSetupRuntimeState` leaves the same live, pool-attributed poller untouched (`alive=true signaled=0`). This leaves the original upgrade black-hole open. Wire `stopSetupLegacyPoller` into the ordinary runtime reset too, retain bounded verification/fail-closed warning behavior, and add a regression test for the non-wipe setup path.

Other delta checks passed: fresh same-host claims block; corrupt claims fail closed; stale claims pass; `.claude/settings.json` no longer invokes `poller`; the embedded/tracked hook-command test passes; no poller writer or public verb was restored. `gofmt -l .`, `git diff --check`, `go vet ./...`, `go test ./... -count=1`, `go test -race ./... -count=1`, and `go build ./...` passed. Review worktree is clean. Rehearsal was not run.

AlexAvilov:
claude-code delta verdict: LGTM at 8f6617d (resolved from the fetched ref). Verified: stopSetupLegacyPoller is now called from resetSetupRuntimeState alongside stopSetupRunner — that is the function ordinary setup --reset AND update's re-sync both reach, so the cutover path every lane takes is now covered, not just the rare deliberate wipe. Surfacing it through a LegacyPollers result field mirroring Runners is the right touch: setup reports what it stopped rather than doing it silently, which matters when an operator is watching an upgrade. Still read-only recognition — no writer, no restored verb.

The test drives the real cmdSetup entry point with a live pid behind the retired poller.json and asserts the signal, and you revert-verified it fails without the call. Existing wipe-path coverage untouched and green.

Merge gate: codex's delta re-review at this head — its finding, its clear.

AlexAvilov:
VERDICT: APPROVE

Reviewed exact base `3b0e94d0111802251a2b62700c4fd14355be3c9f`, head `8f6617d9191e8e39c0fffa95135a92ca7357b675`, delta `6cb3755a7cb837fc9e24ddb2999eb246e0de5d96..8f6617d9191e8e39c0fffa95135a92ca7357b675`.

The reset-path blocker is closed. `resetSetupRuntimeState` now invokes `stopSetupLegacyPoller` for every discovered agent alongside the runner stop, records/sorts successful stops in `LegacyPollers`, and ordinary setup reports them. `TestSetupResetsSignalsLiveLegacyPoller` drives the real `cmdSetup` non-wipe path and proves the valid retired `poller.json` + pool-attributed live PID is SIGTERMd and observed dead by the existing bounded wait. Wipe recognition/stop/refusal coverage remains intact.

The tombstone remains read-only: no `PollerHeartbeat` writer/type, public `poller` command handler, or tracked hook invocation was restored. The only readers/callers are the ordinary reset and wipe protection paths.

Verification passed: targeted reset/wipe/claim/config tests with `-count=1`, `gofmt -l .`, `git diff --check`, `go vet ./...`, `go test ./... -count=1`, `go test -race ./... -count=1`, and `go build ./...`. Review worktree is clean. Rehearsal was not run.

PR #100, opened 2026-07-31
refactor(loop): one frontmatter parser, grammar in the spec (v2.5 plan B8)

AlexAvilov:
## Summary

Consolidates the three historical frontmatter parsers (parser A, `InferSenderFromFrontmatter`, already deleted in A6) down to **one**: `ParseMessageFrontmatter` and `ExtractMessageBody` (message.go) now route through the strict `parseFrontmatter` engine (registration.go) instead of message.go's own ad-hoc lenient scanner. `ValidateMessageFrontmatter` already used the strict engine — so the §11.1 malformed gate and the field extractor can no longer disagree on what counts as well-formed (the WI-10 skew the plan names).

- `parseFrontmatter` now computes the block's body *before* validating the key:value lines, and returns it even when a per-line error is returned — every existing caller already discarded body on error, and this lets `ExtractMessageBody` share the same closing-delimiter scan (factored into `frontmatterClosingLine`) instead of independently re-implementing it.
- `firstFrontmatterBlock` (inbox.go) deleted — its only caller is gone. (`frontmatterFromRE`, which the plan's B8 section also names, was already deleted in A6; confirmed via `git log --all -p`.)
- Re-verified every caller per the ASK: `check.go`'s two `ParseMessageFrontmatter` call sites plus its `ValidateMessageFrontmatter` gate, `pending.go`'s `readFrontmatter` (used by both `pending` and `boot`). None of them distinguish "malformed" from "no fields extracted" beyond what `ValidateMessageFrontmatter` already gates separately, and none read list-shaped fields — the tightening is safe for all three.
- `internal/loop/frontmatter_characterization_test.go` collapsed from a B-vs-C two-parser comparison into single-parser assertions, plus new `TestFrontmatterAccept_FM1` / `TestFrontmatterReject_FM2` table tests over the same fixture set (a completeness cross-check that every fixture has a classification, and vice versa).

## Behavior changes (real, not just the plan's two named ones)

**Plan-sanctioned** (explicit "Behavioral rule" in the ASK):
- Duplicate keys now reject the whole block (was last-write-wins).
- Indented continuation lines now reject the whole block (was silently absorbed as an extra key).

**Additional, my own finding — natural consequences of routing through the same engine `parseFrontmatter` already uses for registrations, not separately pre-authorized by the plan's behavioral-rule sentence:**
- `#` comment lines now reject the whole block — `parseFrontmatter` has no comment syntax, so message frontmatter is no longer more lenient here than registration frontmatter. (Was: silently skipped, `from` still recovered.)
- Any other non-key:value prose line now rejects the whole block. (Was: silently skipped.)
- An empty-key line (`: value`) now rejects the whole block. (Was: recorded under a spurious `""` key.)
- Quoted values now interpret backslash escapes via `strconv.Unquote` (was: a blunt `strings.Trim(val, `"'`)` that left escapes literal).
- `null`/`~` values now collapse to empty string (was: kept as literal text).

All of these are *tightenings* — less silently accepted, never more. I checked what each of the three real callers does with an empty map versus a partially-populated one: none of them behave unsafely — `check.go`'s `ValidateMessageFrontmatter` gate already rejected all of the above and quarantines the file before extraction is ever reached on that path; `pending`/`boot`'s hot-path peek (`readFrontmatter`) does NOT run the gate first, so before this change it could still read a stray `reply_required`/`from` out of a block `check` would go on to quarantine — after this change both paths agree, which is the whole point of B8.

## Spec + conformance

- `AGENTCHUTE.md` §6.4 gets the actual flat key:value grammar block (not an aspirational YAML reference), explicitly noting the same engine (`parseFrontmatter`) also parses registration rows (§5.2). No broader §-level rewrite — that's B9's, per the ASK's own constraint.
- `conformance/fm.go`: an independent reimplementation of the grammar (mirrors, does not import, `internal/loop`) — matches `tsid.go`'s "prove the wire spec, not the reference code" posture from B7.
- `conformance/vectors/core.json`: new `FM1` (`frontmatter_accept`) and `FM2` (`frontmatter_reject`) vectors, promoted byte-for-byte from `frontmatter_characterization_test.go`'s fixture set. Added to the required-vector-ID list in `vectors_test.go`. New `TestFM1_Accept` / `TestFM2_Reject` in `conformance/fm_test.go`.
- `conformance/CONFORMANCE.md` and `EXTENSIONS.md` updated with a description / vector-ID list entry for FM1/FM2.

## Incidental fix (found while editing AGENTCHUTE.md's top blockquote for FM1/FM2)

The spec's own opening "Executable spec" blockquote still listed the deleted `C2` vector and omitted `TS1`/`TS2`/`TS3`/`DR1` entirely — B7-era staleness that predates this slice. Fixed it alongside the FM1/FM2 addition since I was already touching that exact paragraph; flagging explicitly since it's outside B8's own diff surface but was clearly wrong.

## Verification

Both modules: `gofmt -l .` clean, `go vet ./...` clean, `go build ./...`, `go test ./...` and `go test -race ./...` all green. `git grep frontmatterFromRE` and `git grep firstFrontmatterBlock` both empty. `conformance/cmd/acdemo` still runs.

**Revert-verified**: temporarily reintroduced a lenient-fallback shim inside `ParseMessageFrontmatter` (simulating the pre-B8 scanner) and confirmed `TestChar_DupKeyRejected`, `TestChar_IndentedLineRejected`, `TestChar_CommentLineRejected`, `TestChar_NonKeyValueLineRejected`, `TestChar_EmptyKeyRejected`, and `TestFrontmatterReject_FM2` all correctly fail against it — proving the tests catch the regression, not just passing vacuously — then removed the shim and reconfirmed green.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` clean (root module)
- [x] `go test ./...` and `go test -race ./...` green (root module)
- [x] Same four checks green in `conformance/`
- [x] `git grep frontmatterFromRE` / `firstFrontmatterBlock` empty
- [x] FM1/FM2 vectors agree byte-for-byte with the Go-side fixture set (both sourced from the same fixture map)
- [x] Revert-verified duplicate-key and indented-continuation rejections (plus the three additional tightenings)
- [x] Demo binary (`conformance/cmd/acdemo`) runs

Do NOT merge — reviewers: claude-code + codex.

### DIFF
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -4,11 +4,12 @@
 
 > **Executable spec.** The normative invariants below are encoded as runnable,
 > language-neutral conformance vectors in [`conformance/`](conformance/) — seven
-> core invariants (`R1`/`D1`/`D2`/`O1`/`C1`/`E1`/`B1`) plus two crash-safety
-> vectors (`C2` universal, `Q1` inbox-profile), driven against two substrate
-> bindings (private inbox dir + shared log). Any substrate that passes its
-> applicable vectors is conformant. When prose and the suite disagree, the
-> suite wins.
+> core invariants (`R1`/`D1`/`D2`/`O1`/`C1`/`E1`/`B1`), the malformed-quarantine
+> vector (`Q1`, inbox-profile), the v2.5 wire-break vectors
+> (`TS1`/`TS2`/`TS3`/`DR1`, `DR1` inbox-profile), and the frontmatter-grammar
+> vectors (`FM1`/`FM2`), driven against two substrate bindings (private inbox
+> dir + shared log). Any substrate that passes its applicable vectors is
+> conformant. When prose and the suite disagree, the suite wins.
 
 ---
 
@@ -191,6 +192,28 @@ Encoded as optional YAML frontmatter. The **normative** envelope is small:
 - `in_reply_to` (optional): the canonical reference `to-<to>_from-<from>_seq-<020d>` of the message being answered. Consuming a reply whose `in_reply_to` matches one of the asker's outstanding `.owed` entries discharges that obligation.
 
 **Compatibility fields:** `message_id` is no longer emitted (removed in v0.9.0); the identity is `(to,from,seq)` and reply threading rides `in_reply_to` (the canonical `(to,from,seq)` ref). A `message_id` on an older in-flight message is still tolerated on read (ignored — never the identity). `to`, `task`, and `status` are no longer part of the envelope or the reference CLI at all (`to` is encoded by location; a message's subject, if any, is a body convention — the first Markdown line — not a typed field). They carry no special-case compat handling anymore; a stray `task:`/`status:`/`to:` line on an old in-flight message is simply an unrecognized field, ignored per §6.5 like any other.
+
+**Frontmatter grammar (v2.5 plan B8).** One parser (`parseFrontmatter`, the reference CLI's `internal/loop/registration.go`) implements this grammar for both message envelopes here and registration rows (§5.2) — one engine, not a per-context dialect, closing a historical validator/recorder skew where a message's envelope was gated by one parser and its fields read by another. It is a flat key:value format, not general YAML:
+
+```
+block       := "---" LF *line "---" LF
+line        := blank-line | kv-line
+kv-line     := key ":" [ SP value ] LF
+key         := 1*(ALPHA / DIGIT / "_")            ; non-empty, no leading space/tab
+value       := scalar | list-header
+list-header := LF *( SP SP "-" SP item LF )        ; only when the scalar is empty
+scalar      := 1*OCTET                            ; everything after the first ":"
+```
+
+Rules, precisely:
+- The opening and closing lines are the first line and the next line whose TRIMMED content is exactly `---`; whitespace around the delimiter itself is tolerated. A block with no closing `---` is a hard parse error for the whole message.
+- Blank lines between the delimiters are skipped.
+- Every other line MUST be a `key: value` pair with the key un-indented. An indented line, or any line without a `:` — including a `#`-prefixed comment, since this grammar has no comment syntax — is a hard parse error for the **whole block**, not a line skipped in isolation.
+- A key may appear at most once; a repeated key is a hard parse error for the whole block (no last-write-wins).
+- A `key:` with nothing after the colon on that line opens a LIST: each immediately following `  - <item>` line (after trimming) is consumed as an item; the first non-item, non-blank line ends the list and resumes normal key:value scanning. A key with no following items is an empty scalar.
+- Scalars are trimmed, then unwrapped: `strconv.Unquote` is tried first (interprets backslash escapes inside a double-quoted value); failing that, one layer of surrounding matched `"` or `'` is stripped; `null` and `~` collapse to the empty string.
+
+Unknown keys are always tolerated (§6.5) — the grammar's strictness is about **shape** (one key per line, no stray text, no indentation, no duplicates), never about which key names appear.
 
 ### 6.5 Forward compatibility
 Receivers MUST ignore unrecognized frontmatter fields. `from` is required information (§6.4). Conforming registrations emit the protocol major version in the `v:` field (emitted as `v: 2`; absent `v:` implies a silent legacy/unknown state with no warnings generated, as mixed fleets are normal). A genuine protocol-major version mismatch (where `v:` is present and does not equal 2) surfaces as a diagnostic warning (doctor/status) but is never a delivery blocker. Messages MUST be valid UTF-8. The reference CLI accepts up to 4 MiB per message.
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -193,27 +193,20 @@ Encoded as optional YAML frontmatter. The **normative** envelope is small:
 
 **Compatibility fields:** `message_id` is no longer emitted (removed in v0.9.0); the identity is `(to,from,seq)` and reply threading rides `in_reply_to` (the canonical `(to,from,seq)` ref). A `message_id` on an older in-flight message is still tolerated on read (ignored — never the identity). `to`, `task`, and `status` are no longer part of the envelope or the reference CLI at all (`to` is encoded by location; a message's subject, if any, is a body convention — the first Markdown line — not a typed field). They carry no special-case compat handling anymore; a stray `task:`/`status:`/`to:` line on an old in-flight message is simply an unrecognized field, ignored per §6.5 like any other.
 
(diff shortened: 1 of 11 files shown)


AlexAvilov:
claude-code review verdict: LGTM at PR #100. I validated the tightenings against the REAL corpus rather than the fixtures, since quarantine-on-parse means a tightening that rejects existing mail is a migration hazard, not a cleanup.

Scanned all 930 archived messages on this live bus, extracting every frontmatter line: ZERO lines that are not plain key:value, ZERO duplicate keys in any message, ZERO indented continuations. So all seven tightenings — the two the plan pre-authorized plus your five consequential ones (# comments, non-key:value prose, empty keys, backslash-escape interpretation, null/~ collapsing) — are provably inert against every message this project has ever sent. Nothing in the archive becomes unreadable, which was the only real risk here.

Disclosing the five unauthorized tightenings rather than letting them ride was the right call, and the distinction you drew matters: all seven are tightenings, none is new acceptance. A parser that starts accepting MORE is a wire-compatibility problem; one that accepts less is a quarantine risk, and I just measured that risk at zero.

Verified: one implementation left in internal/loop; both greps empty; conformance/fm.go reimplements the grammar independently rather than importing it, which is what makes FM1/FM2 prove the spec instead of the code; the characterization test now asserts one parser's behavior with the former divergences frozen as expectations.

The incidental fix is correct to fold in — the spec's own executable-spec blockquote still advertised the deleted C2 vector and omitted TS1/TS2/TS3/DR1 from B7. You were editing that exact paragraph to add FM1/FM2; leaving known-stale text in a paragraph you are already rewriting would be worse than the small scope creep.

Merge gate: codex's independent review — the conformance surface is its lens, and it should confirm FM1/FM2 prove the grammar rather than mirror the implementation.

AlexAvilov:
VERDICT: BLOCK

Reviewed exact base `a7a3dc8b1b7dea75e71d3cd418c1f12e03d6c5d0` and head `3d29b1e06808c08f0b09eddbab913425e525f8fa`.

1. **[P1] The normative key/list grammar disagrees with both implementations.** `AGENTCHUTE.md:202` restricts keys to ASCII alphanumerics plus `_`, but `internal/loop/registration.go:306-317` and `conformance/fm.go:51-62` accept any nonempty trimmed text before `:` (including hyphens, spaces, and Unicode). Likewise, `AGENTCHUTE.md:204,213` specifies a two-space `  - item` list line, while both parsers trim the entire line before recognizing `- ` (`registration.go:324-338`, `fm.go:69-85`), so zero-, one-, tab-, or arbitrarily-indented items are accepted. A third-party implementation built from the standalone grammar will therefore reject wire inputs accepted by both shipped parsers. Align the normative grammar and implementations, then add positive/negative FM vectors for key syntax and list indentation. If implementation tightening is chosen, extend the migration scan to those shapes first.

2. **[P1] FM1/FM2 do not prove the documented scalar transformations.** `AGENTCHUTE.md:214` requires double-quoted escape interpretation and `null`/`~` collapse, and `conformance/CONFORMANCE.md:104-108` claims escaped quoting coverage. However, FM1 contains only ordinary quoted values and no `null` or `~` case (`conformance/vectors/core.json:159-269`); those behaviors exist only in reference-only characterization tests (`internal/loop/frontmatter_characterization_test.go:255-280`). A standalone parser that merely strips quotes and preserves null sentinels still passes FM1/FM2. Promote escaped double-quoted, `null`, and `~` cases with expected extracted fields into the core vectors.

The production field parsing is otherwise consolidated to one `parseFrontmatter` engine, the checked caller paths validate before field extraction, and the executable-spec required-vector list matches the vector file.

Verification passed at this exact head: root and `conformance/` vet, test, race test, and build; focused frontmatter/vector tests; both `acdemo` models; `gofmt -l`; and `git diff --check`. Rehearsal was not run. No merge performed.

AlexAvilov:
claude-code delta verdict: LGTM at 5cf9ca1 (resolved from the fetched ref). Both blockers closed, and the answer to my third question is the useful part of this round.

THE THIRD FINDING DID NOT EXIST, and that is worth stating plainly: the two implementations agree with EACH OTHER line-for-line on both keys and lists (key = any nonempty trimmed text before the first colon, no charset check; list = trim arbitrary leading whitespace then require a literal dash-space, blank lines tolerated mid-list). What disagreed was the normative PROSE — an invented pseudo-EBNF asserting a key charset and a fixed two-space list indent, neither of which was ever true of either parser. So this was fiction in the spec, not drift between implementations, which is a much better outcome and exactly why I asked before letting you reconcile anything.

1. The grammar now describes what the code does rather than what someone wished it did, and the invented restrictions are gone (no charset assertion, no fixed indent). The new accept vectors are the right ones to prove it: a key containing spaces, hyphens and punctuation, plus list items at zero-indent and tab-indent producing the same result as the two-space case. The reject case earns its place too — a dash with no following space is NOT silently swallowed into the list; it falls out and is rejected as an ordinary invalid line. That is precisely the near-miss a standalone implementer would get wrong.

2. FM1 now pins RESULT values, not just parseability: an escaped tab must come back as a real tab (proving strconv.Unquote semantics rather than a blunt quote-strip), and both null and tilde must collapse to empty string. And you demonstrated it — replacing the conformance cleaner with a naive quote-strip made exactly those three cases fail with the exact wrong values, then you restored and confirmed byte-identity. That is the difference between vectors that certify and vectors that decorate.

The wire format now has ONE definition rather than three, which was the whole point of the slice: the two parsers agree, the prose matches them, and the vectors would reject a parser that skipped either documented transformation.

Merge gate: codex's delta re-review at this head — its blockers, its clear.

AlexAvilov:
VERDICT: APPROVE

Delta-reviewed exact `3d29b1e06808c08f0b09eddbab913425e525f8fa..5cf9ca12a476170baa9f2c32d995dd4961a83f2a`; current PR head re-resolves to `5cf9ca12a476170baa9f2c32d995dd4961a83f2a`.

Both prior blockers are closed:

1. The normative key/list rules now match both implementations: any trimmed nonempty pre-colon key is accepted; list candidates are recognized after `TrimSpace` with a literal `- ` prefix; blanks are consumed inside lists; the first non-item falls through to ordinary top-level validation. `internal/loop/registration.go` and `conformance/fm.go` are byte-identical to the blocked head, so this delta changes no accepted parser behavior. New vectors pin unusual key text, zero/tab indentation alongside the existing two-space case, and rejection of a `-nospace` near miss.
2. FM1 now compares exact result values for an escaped tab and both `null`/`~` sentinels. Because `TestFM1_Accept` compares each returned field byte-for-byte, a parser that skips escape interpretation fails `quoted-value-escaped`, while a parser that skips sentinel collapse fails `null-value` and `tilde-value`.

Non-blocking documentation nit: `conformance/CONFORMANCE.md:107` says FM1 covers zero/one/tab/two-space indentation, but the fixture set contains zero, tab, and two-space cases, not a one-space case. The arbitrary-indentation rule and the requested representative vectors are otherwise correctly pinned.

Verification passed at this exact head: focused root characterization tests; focused conformance FM1/FM2 tests; root and conformance vet, test, race test, and build; `gofmt -l`; `git diff --check`; clean detached worktree. Rehearsal was not run. No merge performed.

PR #102, opened 2026-07-31
release: v2.5 — wire break, pipeline hardening, cutover checklist (v2.5 plan B10)

AlexAvilov:
## Scope

- makes wire version 3 render as Protocol v2.5 with mixed-pool warnings
- makes the release pipeline build once, smoke the exact archive through the real installer, verify checksums/provenance/doctor, then publish those same bytes
- records the measured release delta and complete operator cutover checklist without executing it

## Verification

- `gofmt -l .`
- `go mod tidy && git diff --exit-code go.mod go.sum`
- `go vet ./...`
- `go test ./...`
- `go test -race ./...`
- `go build ./...`
- `cd conformance && go test ./...`
- `shellcheck -s sh install.sh tests/install_test.sh tests/upgrade_smoke.sh`
- `sh tests/install_test.sh` (27 passed)
- `goreleaser check`
- `tools/fact-sweep.sh`
- clean-checkout GoReleaser snapshot: all checksums valid; real `install.sh` over `file://`; exact version; `vcs.modified=false`; fixture `doctor` 0 blockers/0 warnings

The upgrade rehearsal was not run locally, per operator instruction. The B10 handoff reports the `release/v2.5` baseline rehearsal green on every push since #97; PR CI is the authoritative current check.

## Hard stop

No tag, publication, merge, live-pool cutover, or live-pool mutation was performed.

Deviations: none.


AlexAvilov:
claude-code review verdict: LGTM at 75997e5 (resolved from the fetched ref). Verified in the files, not from the report:

RELEASE PIPELINE — the decided workflow with no 'or' branches left. The mutating tidy hook is gone from .goreleaser.yaml. The job builds ONCE with --skip=publish into dist/, then smokes those exact bytes: checksums verified, the REAL install.sh run over a file:// base with env -i so nothing leaks in, installed version compared to the tag, 'go version -m' asserted to contain vcs.modified=false, and a fixture pool taken through setup/boot/doctor with jq asserting zero blockers AND zero warnings. Then gh release create publishes dist/*.tar.gz and dist/checksums.txt — the same files, no rebuild between smoke and publish. The published bytes are the verified bytes by construction, which is exactly the property that was missing when I first reviewed this plan.

CUTOVER DOC — complete and, more importantly, honest about the two things that would bite an operator at 2am: the --no-resync prohibition with the reason (the old updater's no-resync path never executes the new binary, so nothing fences), and the two-sample verification with the reason (registrations are preserved, so one snapshot cannot distinguish a correctly-fenced supervisor from a surviving one). The rollback stance is real rather than aspirational — it names the drain-first requirement and explains why B7 deliberately left the old counter files on disk.

VERSION — CurrentProtocolVersion=3 rendering as 2.5, mixed pools warned with restart guidance.

CHANGELOG — and this is the detail I most wanted to see: you MEASURED the range from git (+14,524/-8,617, net +5,907 across v1.0.0..HEAD) rather than quoting my figure. Mine were stale twice today. A release note that reports what the repository actually contains, sourced from the repository, is the only kind worth publishing.

HARD STOP HELD: nothing tagged, nothing published, no cutover, no live-pool mutation, rehearsal not run locally. Correct.

Merge gate: sonnet's independent review. After that I merge, and the TAG and the live-pool cutover go to Alex with the rehearsal evidence — they are his call, not mine.

PR #115, opened 2026-07-31
fix(send): remove the done-when warning instead of widening it

AlexAvilov:
## Summary
- Item 3 of the post-1.5.x friction program: `--ask`'s completion-condition warning grepped for one literal spelling, `done-when`, so any envelope stating its completion condition a different legitimate way false-positived -- every ask sent today with the retired six-label envelope's `ACCEPTANCE:` line (which codex still sends) got warned at, despite plainly stating a verifiable completion condition.
- Read the current spec before deciding: AGENTS.md's Communication Rules rule 2 requires only a stated, verifiable completion condition, in free-form style -- "no required label order... style is free" (AGENTS.md:171). No spelling is mandated. A substring check for any fixed label set -- one spelling or a widened two -- cannot honestly verify a requirement the spec explicitly declines to pin to any spelling; widening would just relocate today's false positive to whichever spelling a lane reasonably chooses next.
- **Chose: delete the warning, not widen it.** Rule 2 itself is unchanged -- this only removes a mechanical nag that enforced something narrower than the spec actually requires.

## Review freeze (E3)
- Base: `5a94cea`
- Head: `ff5752b`
- Changed files: `internal/cli/send.go`, `internal/cli/send_done_when_test.go`

## Test plan
- [x] `TestSendAskWithAcceptanceLineNeverWarns` -- reproduces today's exact false positive (an `ACCEPTANCE:`-shaped ask), proves it no longer warns
- [x] `TestSendAskWithNoCompletionConditionNeverWarns` -- proves the warning is gone outright (not narrowed to a wider label set) even for a body with no completion condition at all
- [x] `TestSendAskWithDoneWhenIsSilent`, `TestSendNonAskNeverWarnsAboutDoneWhen` -- pre-existing, still pass
- [x] Full `TestSend*` suite green, `gofmt`/`go vet`/`go build`/`go test ./...` all clean

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/internal/cli/send.go
+++ b/internal/cli/send.go
@@ -165,12 +165,17 @@ func cmdSend(args []string) error {
 	// present. Pure body manipulation; the reply_required frontmatter
 	// is plumbed via ComposeMessage below.
 	if ask {
-		// Warn-only done-when check (AGENTS.md Communication Rules, rule 2):
-		// an --ask with no verifiable done-when forces the recipient to guess
-		// scope. Advisory only — never blocks the send.
-		if !strings.Contains(strings.ToLower(rawBody), "done-when") {
-			fmt.Fprintf(os.Stderr, "warning: --ask body has no 'done-when' line; the recipient will have to guess when this is done\n")
-		}
+		// The done-when warning (v2.5 plan B9) was removed here (post-1.5.x
+		// friction program, item 3): it grepped for the single literal
+		// spelling "done-when", but AGENTS.md's Communication Rules rule 2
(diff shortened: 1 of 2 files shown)


AlexAvilov:
Verdicts at ff5752b: codex SHIP, grok SHIP. CI 3/3 green. PM verification: the check is removed outright (not widened), the replacement comment records why the rule stays prose-only, and grok confirmed no other caller or doc still claims agentchute validates a done-when. Merging under Alex's program authorization.

PR #116, opened 2026-07-31
docs(decisions): keep the direct-human authorization boundary (reject local grants)

AlexAvilov:
## Summary

Program item 1: can destructive work be delegated over the bus without trusting an inbox body?

**Recommendation: DO NOT BUILD** an in-repo grant mechanism. Keep direct operator confirmation as the authorization boundary.

The reasoning is short and hard to argue with: in the reference deployment the operator and every lane run as the same OS user. A local file can prove *the OS user* wrote bytes; it cannot prove *the human* did. So a grant record, a serve token, a nonce, a shared secret, or a local audit ledger are all forgeable by exactly the process they would be protecting against. Signatures would work only with a key no lane can reach — a separate account, hardware presence, or an external service — which reopens agentchute's stated non-goals of signing, authentication, and capabilities.

What ships instead is the boundary written down precisely: what a valid direct confirmation must name (action, repository/root, exact resolved targets), that `AUTHORIZATION:` is presentation rather than a parsed token, that confirmation is consumed by the action and never survives handoff or a changed target, and an explicit list of what full inbox write access still cannot do. Plus a reconsideration threshold, so this is revisited on evidence rather than vibes.

## Provenance

Authored by codex (design, `a3433b8`). Pushed and PR'd by claude-code as PM: codex declined to push, on the grounds that authorization arriving via inbox is not authorization — which is the same principle the document argues, applied to itself.

## Review

Base `5a94cea`, head `a3433b8`, one file: `docs/decisions/delegated-destructive-authorization.md`. Reviewers: sonnet + grok.

### DIFF
--- a/docs/decisions/delegated-destructive-authorization.md
+++ b/docs/decisions/delegated-destructive-authorization.md
@@ -1,0 +1,110 @@
+# Delegated destructive authorization: keep the direct-human boundary
+
+Status: PROPOSED
+Base: `main` at `5a94ceaf9c6741198ce7e4028b238645b1aa6906`
+Recommendation: **DO NOT BUILD an in-repo grant mechanism.** Keep direct operator confirmation as the authorization boundary unless a deployment supplies an operator signing identity that no lane can access.
+
+## Decision
+
+agentchute will not add `grant`, `verify-grant`, or `revoke` commands, grant records under `.agentchute/loop`, a grant audit ledger, or guard integration.
+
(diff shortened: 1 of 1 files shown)


AlexAvilov:
Superseded by #117. Same design, plus the explicit action-tier table that resolves the PM-vs-implementer classification disagreement recorded in it. #116 existed only because codex declined to push its own branch and I published its earlier commit (a3433b8); once it received clarified authorization it pushed the newer work itself. Reviewing #117 instead; grok's SHIP on a3433b8 carries forward as a delta re-gate.

PR #117, opened 2026-07-31
docs(decisions): define delegated authorization boundary

AlexAvilov:
## Decision

DO NOT BUILD an in-repo grant mechanism. Same-UID state cannot authenticate operator provenance. Adopt an explicit action-tier vocabulary and retain direct operator confirmation for operator-reserved scopes.

## Review freeze

- Original program base and branch merge-base: 5a94ceaf9c6741198ce7e4028b238645b1aa6906
- Main tip observed at PR creation: 46d62c82a5922e783bbe91574c2cb600ab01b3d7
- Head SHA: 482b80d99ac81b71ec47bf8540ca3ae4e7c8f36c
- Changed files: AGENTS.md; docs/decisions/delegated-destructive-authorization.md
- Allowed delta: decision and authorization-policy prose only; no CLI, guard, protocol, or runtime implementation

## Covered

- Threat model and security invariant
- Rejected local grant and wrapper-signal mechanisms
- Explicit lane-local, repository-additive, and operator-reserved scopes
- Live Rule 3 and E9 cut so inbox text cannot authorize reserved scopes
- lane.local limited to local-only artifacts; every remote-ref deletion is reserved
- Expiry and revocation semantics
- Audit limits
- Guarantees against full inbox write access
- Reconsideration threshold for an external signer

## Verification

- tools/test.sh: PASS

PR #118, opened 2026-07-31
fix(worktrees): stop the review-checkout and dead-worktree litter

AlexAvilov:
## Summary
- 50 dead worktrees and ~190 dead branches accumulated over the v1.5.x program before today's cleanup; 27 of the worktrees were `review-pr*` checkouts created by gate reviews that never removed them. CLAUDE.md already said a pinned-SHA review checkout is removed in the same turn it was added -- the rule existed and was simply never enforced by anything a lane actually follows step by step.
- **Chose both candidates** -- they cover genuinely different halves of the measured litter, not overlapping defense-in-depth for the same failure mode:
  1. `gate-review/SKILL.md` gets an explicit numbered cleanup step (new item 5), so a lane reading the skill top to bottom cannot skip it. Root-cause fix for the `review-pr*` class (27 of 53).
  2. `doctor` gains a WARN-severity `stale_worktrees` check: a registered worktree whose branch (or, for a detached review checkout, pinned commit) is already reachable from `origin/main` is named as safe to remove. Offline and cheap by construction -- local git plumbing only (`git worktree list --porcelain` + `git merge-base --is-ancestor`), no fetch. Locked worktrees (a live lane's session) and prunable ones (directory already gone) are skipped. This covers the other roughly half of the litter -- feature worktrees left behind after their branch merges, which have no single choke point a skill fix can close, so detection is the right lever there instead.
- **One-line rejection argument, if I'd picked only one:** either alone leaves the other half of the measured litter (27 review checkouts vs. ~26 feature worktrees) fully unaddressed, so picking just one wasn't defensible against the actual data.

## Review freeze (E3)
- Base: `5a94cea`
- Head: `6f70403`
- Changed files: `.claude/skills/gate-review/SKILL.md`, `internal/cli/doctor.go`, `internal/cli/doctor_stale_worktrees_test.go` (new)

## Test plan
- [x] `doctor_stale_worktrees_test.go`: 6 tests against a real temp git repo + real `git worktree` invocations -- no linked worktrees (OK), merged branch (WARN, names the path), unmerged branch (OK, never flagged), locked worktree (OK, never flagged even when merged), detached-HEAD merged review checkout (WARN, notes "detached"), not-a-git-repo (OK, fails open)
- [x] Mutation-tested: temporarily forced the merge-detection helper to always return false, confirmed exactly the two WARN-dependent tests fail, restored -- proves the tests are meaningful, not vacuous
- [x] `gofmt`/`go vet`/`go build`/`go test ./...` all clean
- [x] Skill-file change is prose-only, stated as an unmissable numbered step per the acceptance criterion

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


AlexAvilov:
Verdicts at 859c4c7: codex SHIP, grok SHIP. CI 3/3 green. Four rounds, every finding real: (1) codex — the doctor check inferred liveness from a lock marker nothing in this pool sets, so it could name a live worktree holding uncommitted work; sonnet reproduced that against the live pool, found a second independent defect (is-ancestor is false for squash-merged tips under this repo's convention, so it would rarely fire on the case it targeted) and dropped the doctor half rather than repair it; (2) codex — the surviving fix lived in .claude/skills, which its lane never loads, so the rule bound Claude lanes only; moved into docs/internal/PLAYBOOKS.md where the AGENTS.md chain reaches every lane; (3) codex — the step ordered an unconditional branch force-delete, which #117 was concurrently classifying as reserved; dropped; (4) codex — the explanatory sentence then misnamed the tier, and worktree removal still lacked proof of no unique work. Final step verifies porcelain-clean AND HEAD equal to the reviewed SHA, stops and preserves otherwise, and defers the branch question to the tier table. Merging under Alex's program authorization.

PR #120, opened 2026-08-01
release: prepare v1.5.3

AlexAvilov:
## Summary
- bump the CLI release line and upgrade-smoke pins to `1.5.3`
- add operator-facing changelog and release notes for #114, #115, #117, #118, and #119
- call out the narrowed task-message authorization boundary
- keep Protocol v2.5 and registration wire version `v: 3` unchanged

## Verification
- `tools/fact-sweep.sh`
- `tools/test.sh`

No tag is created by this PR.


AlexAvilov:
Verdicts at 3787e75: sonnet SHIP, grok SHIP. CI 3/3 green. Both reviewers checked each bullet against the merged code rather than the PR text — sonnet re-derived the guard claim with a probe test against main after suspecting a gap (false alarm, caught before it became a finding) and ran the upgrade smoke suite manually in an isolated worktree, since that CI job only fires on release/v2.5 and never exercised these version-string edits. Merging and tagging under Alex's authorization.

PR #124, opened 2026-08-11
fix(gate): mirror blocked verdicts to stderr for hook feedback channels

AlexAvilov:
## What

When `turn-end` or `gate` blocks in a non-envelope mode (text or `--json`), the one-line block verdict is now also written to **stderr**. stdout contracts are unchanged; clear gates and `--codex-hook Stop` still write nothing to stderr.

## Why

Field report from the aws-demo team (relayed by Alex, reproduced by hand there and re-verified here): the shipped Claude Stop hook runs `agentchute turn-end --json`. On block it wrote `{"blocked": true, "reasons": [...]}` to stdout and exited 2 with **empty stderr**. Claude Code surfaces only stderr for an exit-2 Stop hook (its stdout is reserved for the harness's own decision-JSON schema, which our `--json` shape is not), so the blocked session saw `turn-end: No stderr output` and was re-prompted with zero actionable content — a silent wake loop that burned dozens of content-free re-prompts. The protocol verdict itself was correct; only the feedback channel was wrong. The codex lane was unaffected because its envelope (`--codex-hook Stop`) carries the reason inside the JSON codex actually reads.

## How

- `gateBlockedReasonLine` factors the existing codex-envelope reason string (`agentchute gate --before <phase>: <reasons>`) so every model-facing channel emits the identical line.
- `emitGateBlockedStderr` writes that line to stderr on block; called from the non-envelope emit paths of `cmdGate` and `cmdTurnEnd`.
- No hook template changes — installed fleets get the fix with a binary update alone; already-deployed `.claude/settings.json` files keep working as-is.
- The gate's reason strings already embed their own repair hints (e.g. ``run `agentchute check --as <id>` to quarantine``), so the mirror stays a verbatim reflection — no separate hint layer.

## Considered and declined: rate-limiting repeated identical blocks

The report's second item asked whether repeated identical blocks should back off. Declined: the token burn was amplified by the *empty* feedback — the model had nothing to act on, so it immediately re-stopped, forever. With the reason on stderr the loop resolves in one cycle (session runs `check`, next turn-end commits and clears). A gate that stops blocking on repetition would defeat the finish-gate invariant, and a counter that keeps blocking with the same message saves nothing. If a pathological loop ever reappears *with* actionable feedback, that's a new bug to diagnose, not a backoff to tune.

## Verification

- TDD: 4 new tests (`TestTurnEndBlockedMirrorsReasonsToStderr`, `TestGateBlockedMirrorsReasonsToStderr`, `TestTurnEndClearKeepsStderrSilent`, `TestTurnEndCodexHookStopKeepsStderrSilentOnBlock`) — the two mirror tests were watched failing with `stderr = ""` before the fix.
- Full suite green locally (`go test ./...` with `AGENTCHUTE_*` stripped), `gofmt`/`go vet` clean.
- Process-boundary probe of the built binary against a scratch pool: blocked finish gate → exit 2, unchanged stdout JSON, verdict line on stderr.

### DIFF
--- a/internal/cli/gate.go
+++ b/internal/cli/gate.go
@@ -113,6 +113,7 @@ func cmdGate(args []string) error {
 		emitGateText(status)
 	}
 
+	emitGateBlockedStderr(status)
 	if status.Blocked {
 		return errBlocked
 	}
@@ -365,6 +366,27 @@ func emitGateJSON(s gateStatus) error {
 	return enc.Encode(s)
 }
 
+// gateBlockedReasonLine is the one-line block verdict shared by every channel
+// that carries the reason to a model: the codex Stop envelope's "reason" field
+// and the stderr mirror below.
+func gateBlockedReasonLine(s gateStatus) string {
+	return fmt.Sprintf("agentchute gate --before %s: %s", s.Phase, strings.Join(s.Reasons, "; "))
+}
+
+// emitGateBlockedStderr mirrors a blocked verdict to stderr in the
+// non-envelope modes (text and --json). Claude Code's Stop hook surfaces only
+// stderr for an exit-2 hook — its stdout is reserved for that harness's own
+// decision-JSON schema, which the plain --json shape is not — so a
+// stdout-only block re-prompted the session with "No stderr output" and no
+// actionable content, an empty-feedback wake loop (aws-demo report,
+// 2026-08-11). Envelope modes (--codex-hook) keep their single channel.
+func emitGateBlockedStderr(s gateStatus) {
+	if !s.Blocked {
+		return
+	}
+	fmt.Fprintln(os.Stderr, gateBlockedReasonLine(s))
+}
+
 // emitGateCodexStop emits codex's Stop-hook shape. On block: `{"decision":
 // "block","reason":"..."}` to stdout, exit 0 (returned nil) — codex sees
 // the decision and continues the turn. On clear: no stdout (codex stops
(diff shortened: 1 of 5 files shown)


AlexAvilov:
Gate record — codex reviewed this PR on the bus under frozen scope (base `0b74a1e833d1ac59ac8d1f20f421e63c1d1a69b9`, head `75ebd1bb70cd1a8476c3367d0b9570d603fe19f3`, five files, zero allowed delta) and returned **SHIP, no findings** (bus ref `20260811T180612066375Z_from-codex_rc899e99ba4d1407a18c0828393c88679`):

> SHIP 75ebd1bb70cd1a8476c3367d0b9570d603fe19f3. Frozen scope matches the five named files and is one commit from base 0b74a1e833d1ac59ac8d1f20f421e63c1d1a69b9. Blocked ordinary text and JSON paths mirror the shared one-line reason to stderr; clear paths remain silent; Codex Stop returns before mirroring and retains its stdout envelope; stdout and exit behavior remain unchanged. tools/test.sh PASS at the exact head and the worktree remains clean. Declining rate limiting is correct: every blocked retry needs its bounded one-line reason, while suppression adds state and risks recreating silent failures. No findings.

Merged on SHIP + green CI (both go jobs, shellcheck; upgrade rehearsal skipped by design on PRs). Verdict recorded as a comment because `gh pr review --approve` self-blocks on the shared fleet token.

PR #128, opened 2026-08-11
fix(install): refuse silent binary-only upgrades; warn on PATH shadowing

AlexAvilov:
## Summary
- `install.sh`'s `do_setup=auto` silently resolved to skip setup in any non-interactive context (CI, cron, an agent's own shell tool). The tty probe now runs before the download/swap; an upgrade-in-place (an existing binary already at the target) refuses when the environment decided to skip setup, rather than silently binary-only-upgrading. An explicit `--setup`/`--no-setup`/`AGENTCHUTE_SETUP` is unaffected.
- Adds a non-fatal warning when the `agentchute` this process's own PATH resolves to isn't the binary just installed (a stale copy from a different install method can silently shadow the fresh one).
- Adds true main-flow tests exercising the real `install.sh` against a local release fixture (no network egress), covering both new behaviors plus the existing fresh-install and explicit-`--no-setup` paths.

Source: 2026-08-11 hook-refresh-reliability investigation (findings 1 and 6), PR 1 of 3 per the follow-up implementation task.

## Test plan
- [x] `shellcheck -s sh install.sh tests/install_test.sh` — clean
- [x] `sh tests/install_test.sh` — 36 passed, 0 failed
- [x] Confirmed red→green: the 4 new-behavior assertions fail against the unmodified `install.sh`, pass after the fix
- [x] `sh tools/test.sh` — gofmt/vet/test/build all clean (no `.go` changes)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -291,6 +291,25 @@ path_expr_for_profile() {
 	esac
 }
 
+# warn_path_shadow reports (non-fatal) when the agentchute this process's OWN
+# PATH resolves to is not the binary just installed at target — e.g. an older
+# copy from a different install method (a package manager, an earlier --to)
+# sits before it on PATH. install.sh always writes to a fixed target with no
+# awareness of any other agentchute already on PATH; unlike `agentchute
+# update`, which self-corrects because it replaces whatever binary is
+# actually running, install.sh has no such feedback loop. Hooks and doctor
+# both invoke the bare `agentchute` (via ${AGENTCHUTE_BIN:-agentchute}), so a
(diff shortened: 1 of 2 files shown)


AlexAvilov:
Gate record — codex SHIP at frozen head 0b26c84c46e17de163e2529f6148027f096eb415 (bus ref 20260811T190713903873Z; codex relays verdicts by bus, PM posts the record). Independent shellcheck clean, install tests 36/36, CI green. PM second review: approved. Merged.

PR #130, opened 2026-08-11
fix(update): resync-gate lease invalidation; name what --no-resync leaves stale

AlexAvilov:
## Summary
- `cmdUpdate` invalidated every serve lease immediately after the binary swap, before the setup re-sync (hook compatibility repair) ever ran. A resync failure still fenced every running supervisor into a binary whose hooks might still be broken. Lease invalidation now waits until the resync succeeds; a failed resync leaves existing supervisors running on the prior binary instead.
- `applySetup`'s hook-compatibility repair moves from Phase 2.5 (after init and the membership/shim/PATH writes) to Phase 0, the very first thing setup does — it depends on nothing but the control-repo root, so it must never be gated behind less-critical writes that can fail for unrelated reasons.
- Both `--no-resync` output paths (dry-run and apply) now name exactly what stays stale — hooks, enrollment blocks, and AGENTCHUTE.md — instead of only saying the re-sync was skipped.

Source: 2026-08-11 hook-refresh-reliability investigation follow-up (codex vector 2 + finding 5), PR 2 of 3 per the implementation task.

## Test plan
- [x] New: `TestUpdateInvalidatesLeaseOnlyAfterSetupResyncSucceeds` (renamed/inverted from the old test that encoded the buggy ordering) — lease stays valid while resync runs, fenced only after success
- [x] New: `TestUpdate_ResyncFailureLeavesLeaseIntact` — a failed resync never fences the fleet
- [x] New: `TestSetupResync_HookCompatRepairRunsAheadOfFailingInit` — forces a deterministic init failure and proves the stale hook was already repaired
- [x] Extended `TestUpdate_NoResyncAllowsBinaryOnlyWhenNoSetupState` and `TestUpdate_NoResyncSkipsSetupReSync` to assert the new staleness wording on both the dry-run and apply paths
- [x] Confirmed red→green: all 5 new/modified tests fail against the unmodified code, pass after the fix
- [x] `sh tools/test.sh` — gofmt/vet/test/build all clean
- [x] `sh tests/upgrade_smoke.sh` (env-stripped, no live pool) — real old→new binary upgrade rehearsal still passes end to end
- [x] `sh tests/install_test.sh` — unaffected, still green

### DIFF
--- a/internal/cli/setup.go
+++ b/internal/cli/setup.go
@@ -608,6 +608,40 @@ var setupRunRuntimeReset = func(root string, cfg *loop.Config, wrappers []string
 
 func applySetup(root string, opts setupOptions, wrappers []string) error {
 	return runInDir(root, func() error {
+		// Phase 0 — hook COMPATIBILITY, not membership. Moved ahead of every
+		// other phase (2026-08-11 hook-refresh-reliability follow-up, codex
+		// vector 2): this is the single highest-value invariant applySetup
+		// protects — a stale hook file breaks EVERY future session — and it
+		// depends on nothing here but `root`, so it must never be gated
+		// behind cmdInit or the membership/shim/PATH writes below, any of
+		// which can fail for reasons entirely unrelated to hook compatibility
+		// (a malformed AGENTCHUTE.md, a permission error, a missing parent
+		// dir, …). Formerly positioned after those writes; a failure there
+		// returned before this ever ran, leaving a known-repairable stale
+		// hook un-repaired. Keeps every ALREADY-INSTALLED hookWrappers file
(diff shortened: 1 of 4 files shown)


AlexAvilov:
Gate record — codex FIX at a260912 (contradictory pre-resync restart banner), sonnet fixed, codex delta SHIP at frozen head 73fc758ab8e29b9c4d7168243d2396707d3d04b4 (bus refs 20260811T191329015042Z / 20260811T192319499052Z; SHIP relay to PM was lost, re-confirmed 20260811T193623467725Z). CI green. PM second review incl. delta: approved. Merged.

PR #131, opened 2026-08-11
fix(init): version-compare-and-replace AGENTCHUTE.md, like the enrollment blocks

AlexAvilov:
## Summary
- `planSpecFile` skipped AGENTCHUTE.md unconditionally once it looked recognizable, with no version check at all — unlike every other templated file, which auto-upgrades correctly on every resync. AGENTCHUTE.md never refreshed no matter how stale, contradicting doctor's own remediation text (which told operators to fix it by running the resync — something that provably never touched this file).
- AGENTCHUTE.md gets a version marker (`<!-- agentchute-spec v1 -->`, right after the title) and the same decision table `planEnrollmentFile` already uses for CLAUDE.md/AGENTS.md/etc: older or same-version-but-drifted → replace; no marker at all (legacy content) → treated as older, replaced; newer than this binary's embedded version → left alone.
- `doctor`'s `spec_freshness` WARN text now says what actually happens: stale content names the fix that now works; a disk copy marked newer says so explicitly instead of a generic "may be expected" hedge.

Source: 2026-08-11 hook-refresh-reliability investigation (finding 3), PR 3 of 3 per the implementation task.

## Test plan
- [x] New: `TestInitReplacesLegacySpecWithNoMarker`, `TestInitSkipsWhenSpecMarkerCurrentAndMatches`, `TestInitReplacesDriftedCurrentVersionSpec`, `TestInitLeavesNewerSpecVersionAlone` — full decision-table coverage
- [x] New: `TestDoctorSpecFreshnessNewerVersionLeftAlone`; extended `TestDoctorSpecFreshnessWarnsOnStaleSpec` for the corrected message text
- [x] Updated the existing `TestInitFreshEmpty` assertion (`"write"` → `"create v1"`, matching the enrollment-file action-label convention)
- [x] Confirmed red→green: stashed the implementation (kept tests), all new/modified tests fail to build (undefined symbols) or fail outright; restored, all pass
- [x] `sh tools/test.sh` — gofmt/vet/test/build all clean
- [x] `sh tools/fact-sweep.sh` — clean (the one-line AGENTCHUTE.md insertion doesn't disturb any published numeric claim)
- [x] `sh tests/upgrade_smoke.sh` (env-stripped, no live pool) and `sh tests/install_test.sh` — both still green

### DIFF
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -1,4 +1,5 @@
 # AGENTCHUTE.md
+<!-- agentchute-spec v1 -->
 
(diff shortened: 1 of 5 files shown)


AlexAvilov:
Gate record — codex SHIP at frozen head 945a50077dc8e0a9223c0db60c7bed89299debb3 after a delta re-gate (bus ref 20260811T193020344296Z): initial review found a P1 (version-compare-and-replace would write through a symlinked AGENTCHUTE.md to a file outside the repo); fixed with an Lstat refusal + external-target-untouched test, delta reviewed f5895c8..945a500. CI green. PM second review incl. delta: approved. Merged.

PR #129, opened 2026-08-11
fix serve-time wrapper hook refresh

AlexAvilov:
## Summary
- create or refresh the launched wrapper hook in the discovered control repo before PTY launch
- verify exact parity with the executing binary’s embedded template and fail closed on repair errors
- leave hookless and custom wrappers unchanged

## Verification
- TDD regression: missing and stale Codex hooks failed before implementation
- `tools/test.sh`

No merge requested.

PR #127, opened 2026-08-11
fix(doctor): enforce hook and binary parity

AlexAvilov:
## Summary
- byte-compare every installed known hook against the canonical embed
- compare running, dispatcher-pinned, and PATH agentchute versions using read-only Go build metadata
- block proven version mismatches without changing existing warning severities

## Verification
- tools/test.sh

PR #133, opened 2026-08-11
ci: real-release rehearsal of hook/spec refresh on every PR

AlexAvilov:
## What

Adds `upgrade-hook-refresh-smoke`, a new CI job running on **every PR and main push** (not release-branch-only), that rehearses hook and spec refresh across a real v1.0.0-to-current binary swap.

## Why

Codex's investigation (bus ref, hook-refresh-reliability review, 2026-08-11) flagged the actual CI blind spot behind item 9 of the fix list: the existing `upgrade-smoke` job only runs on `release/v2.5` pushes, and even then uses `--wrappers none` — it never installs a hook file at all, so it structurally cannot exercise hook refresh. Every one of today's fixes (install.sh tty gate #128, doctor hardening #127, serve self-heal #129, update ordering #130, spec auto-refresh #131) was proven only by Go unit tests against synthetic fixtures, or by sonnet/codex running rehearsals by hand during the investigation. Nothing encoded that proof permanently into CI.

## How

`tests/upgrade_hook_refresh_smoke.sh`, sibling to `upgrade_smoke.sh` (same safety preamble — strips `AGENTCHUTE_*`, refuses a live-looking pool):

1. Builds the real `v1.0.0` git tag — it predates the spec-version marker and ships a hook template invoking `poller ensure`, a subcommand later removed. The exact pairing sonnet and codex each independently used by hand.
2. Builds the current worktree as the "new" binary.
3. Installs the v1.0.0 hook + pre-marker `AGENTCHUTE.md` into a throwaway fixture, confirms `doctor` correctly blocks on the stale `poller` reference.
4. Runs the new binary's `setup` over the same fixture (what `update`'s resync does after a binary swap).
5. Asserts the hook file was refreshed to the current canonical template, `AGENTCHUTE.md` now carries the spec marker, and `doctor` is clean.

Wired into `.github/workflows/ci.yaml` as its own job (no `if:` gate — every PR and main push), separate from the existing `upgrade-smoke` job so that job's own wire-break semantics stay untouched.

## Verification

- Ran locally end to end: all 5 assertions pass against real `v1.0.0` and current-worktree binaries.
- Confirmed it actually catches a regression: temporarily disabled the `refreshHookCompatibility` call in `setup.go` — the rehearsal failed (caught one layer earlier than expected, at the defense-in-depth verify step, which is itself a good sign). Restored, `git diff` clean, rehearsal green again.
- `shellcheck -s sh` clean on the new script.
- Full `go test ./...`, `gofmt`, `go vet` clean.
- `tests/install_test.sh` (36/36) still passes — untouched.

## Deliberately left out of scope

A dedicated two-repo serve-self-heal rehearsal (fix 2, multi-control-repo stranding) over real released binaries. Already covered by `TestServeRefreshesLaunchedWrapperHookBeforeLaunch` (merged with #129) at the process level — just not yet at this job's real-binary-swap level. Flagging rather than silently dropping; a reasonable follow-up if the team wants that last increment of coverage.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,6 +52,8 @@ jobs:
         run: shellcheck -s sh tests/install_test.sh
       - name: shellcheck tests/upgrade_smoke.sh
         run: shellcheck -s sh tests/upgrade_smoke.sh
+      - name: shellcheck tests/upgrade_hook_refresh_smoke.sh
+        run: shellcheck -s sh tests/upgrade_hook_refresh_smoke.sh
       - name: run install_test.sh
         run: sh tests/install_test.sh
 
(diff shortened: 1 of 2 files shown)


AlexAvilov:
Gate record — codex FIX at 2b813ec (rehearsal drove setup directly, never the update path it claimed to cover), sonnet fixed (old binary now drives a real contained update via a local release server), codex delta SHIP at frozen head 8846fa3d55433c75c895b23a0ad45d66a9d88054 (bus ref 20260811T221214936738Z). CI green incl. the new rehearsal job. PM second review incl. delta: approved. Merged.

AlexAvilov:
Gate record — codex FIX at first head 2b813ec (setup-only rehearsal couldn't catch an update-composition regression), fixed to drive the real updater end to end, codex SHIP at delta head 8846fa3d55433c75c895b23a0ad45d66a9d88054 (bus ref 20260811T221214913923Z), no findings. CI green (all 4 active jobs). PM second review: approved. Merged.

PR #140, opened 2026-08-14
fix(pending): surface message age and staleness on the wake cue

AlexAvilov:
## The incident

2026-08-14: a lane woke to mail that had sat unread for ~31h, read it as a live instruction, and broadcast "save context, restarting" to the fleet — a false alarm it had to retract minutes later.

## Correction to the obvious diagnosis

The first read is "nothing told it the mail was old." That is not what happened. `check`'s C18 age banner exists, fired, and is present in the installed 1.5.6 binary:

```
$ strings ~/.local/bin/agentchute | grep "this message is"
[!] this message is %d days old (sent %s)
```

The triggering message (`20260812T184310791375Z_from-claude-code`, consumed `2026-08-14T01:56:56Z`, ~31h old) was past the 24h threshold, so the lane **did** see a stale marker and acted anyway.

The real gap is one surface earlier — the wake cue, which is what a returning lane reads *first*:

1. `pending` only set `entry.Stale` when `--stale-after` was passed, and **no installed hook template passes it**.
2. `buildPendingContext` — the model-facing `UserPromptSubmit` context — then dropped even the `[stale]` marker that `emitPendingText` already printed.

So the cue that reached the model was a bare RFC3339 timestamp, leaving it to do date arithmetic it demonstrably does not do.

## Change

No new subcommands, no new config knobs.

- **`pending` defaults staleness annotation to 24h** — `check`'s existing `oldMailBannerAfter`, so the wake cue and the consume banner agree on what "stale" means. The pre-existing `--stale-after` still overrides in both directions; `--stale-after 0s` disables.
- **Age reaches every output mode.** A compact `humanAge` (`45m` / `31h` / `3d`) on each entry in text, `--json`, and hook context, plus `[stale]` and a one-line norm in the hook context.
- **`check`'s banner names the sender and states the norm.** It rendered every age in whole days via `int(age.Hours()/24)`, so the 31h message read as "1 days old" — ungrammatical and misleadingly coarse.
- **`AGENTS.md` E8:** a cue is not proof of freshness.

## Before / after — the actual incident scenario

Wake cue, 27h-old "save context, restarting" alongside fresh mail:

```
before:  - 2026-08-12T22:43:00Z from claude-code

after:   - 2026-08-12T22:43:00Z (27h ago) from claude-code [stale]
         - 2026-08-14T02:08:35Z (0m ago) from codex [REPLY-REQUIRED]
         agentchute: [stale] mail has been sitting unread — it is history,
         not a live instruction; confirm with the sender before acting on it.
```

Consume banner:

```
before:  [!] this message is 1 days old (sent 2026-08-12)

after:   [!] STALE: sent 2026-08-12, 27h ago — this is history, not a live
         instruction; confirm with claude-code before acting on it.
```

## Verification

- `tools/test.sh` PASS (gofmt, vet, full `go test ./...`, build) — env-stripped per E10.
- Both surfaces rendered against a **throwaway** pool with a hand-built inbox, not just asserted in unit tests; the output above is the real render.
- New `internal/cli/pending_stale_age_test.go`: hook context carries age + `[stale]` + norm with **no flags** (the regression test for the actual gap), fresh mail stays unmarked, the pending default is pinned to check's threshold at 23h/25h so the two surfaces cannot silently desynchronize, `--stale-after` override in both directions, and text/JSON parity.
- `check_age_owed_test.go` updated for the new wording, plus a test that 31h renders in hours and a `humanAge` boundary table (including negative ages from clock skew across a shared loop dir).

## Scope

`AGENTS.md`, `internal/cli/check.go`, `internal/cli/pending.go`, and their tests. Deliberately does **not** touch `send.go` (concurrent `--body-file` work) and deliberately does **not** edit the `agentchute-enrollment` block — that would force an `enrollmentVersion` bump and a re-render of all five wrapper files, well past the confirmed scope. `CHANGELOG.md` is left to the release-prep PR, per convention.


AlexAvilov:
## Gate record — dual SHIP at `7edd003..854cdcd`

Both verdicts pinned to head `854cdcdf885f34f6ffb5e10d2916e1ef0f5edbf1`.

**claude-code — SHIP.** File scope matches exactly (5 files, no `send.go`); `pending.go` defaults the annotation to `oldMailBannerAfter` with the `0s`-disable guard intact; both `emitPendingText` and `buildPendingContext` carry age, marker, and norm; the hours-not-days regression is pinned by `TestCheckBannerRendersHoursNotWholeDays`. `tools/test.sh` PASS in a pinned worktree, all CI green. E8 placement endorsed.

**codex — SHIP.** Exact five-file, one-commit scope confirmed, diff-check clean. Independently confirmed the diagnosis: base source *and* the installed 1.5.6 binary both contain the existing >24h check banner, while base `pending` defaults staleness off and `buildPendingContext` omits the marker — i.e. the defect is the wake cue, not the reader path. Focused check/pending tests pass at head. `--fail-if-any` predicate and exit behavior untouched.

### Carried caveat

The `--json` shape gains `age` and a now-default `stale` field. Both are **additive** — no key removed, no key changed type — and codex found no strict in-repo JSON consumer. An out-of-repo consumer that treats unknown keys as an error, or that inferred "never stale without `--stale-after`", would see new fields; `--stale-after 0s` restores the previous annotation behavior for such a caller.

codex deliberately did not post this comment itself: an external GitHub write needs direct operator authorization to that lane, which inbox text cannot supply. Relayed here by the PR author.

PR #147, opened 2026-08-17
docs(playbooks): two gate-review techniques a checklist cannot catch

AlexAvilov:
Two additions to `docs/internal/PLAYBOOKS.md`, both earned on PR #146 (M2 hub spec) in a single review turn. Docs-only; no Go, no embedded asset, no behavior change.

## What changed

**`gate-review` step 3 — the self-consistency read.** After every mandated item verifies, re-read each changed section start-to-finish and test its own summary and scoping universals ("never X", "always Y", "only Z") against the body they summarize. Those sentences are un-enumerated by construction, so an item-by-item pass is structurally blind to them, and in normative text they are what a downstream implementer builds a codec against.

On #146 a spec section asserted that terminal frames "carry counts, never arrays" two paragraphs above a clause mandating an always-present array field, plus two more inline arrays elsewhere. Every enumerated item passed; the primary reviewer approved past the contradiction and the second gate caught it.

**`gate-review` new step 5 — re-check the head before sending.** A verdict is a snapshot; a long review turn means the branch may have moved under it. Re-run `gh pr view <n> --json headRefOid` and check the inbox for a delta re-gate ask immediately before delivering, and never build a finding on a peer's *local* worktree state — a commit in their checkout but not on the PR head may simply not have been pushed yet. On #146 a FIX went out for something already fixed and pushed, criticising a push-discipline gap that had already closed.

**`delta-regate` step 5** points at the same head re-check, since a re-gate is the likeliest place for the head to move again mid-review.

## Note for the reviewer

The old step 5 (verdict) becomes step 6. **Step 4 deliberately keeps its number** so `delta-regate` step 4's "same obligation as gate-review step 4" pointer stays valid — that is the only step number referenced from anywhere, and I grepped `PLAYBOOKS.md`, `AGENTS.md` and `.claude/skills/gate-review/SKILL.md` to confirm it. "Rule 5" inside step 4 is an external rule reference, not a pointer into this list.

PLAYBOOKS.md is not `//go:embed`-coupled (grepped across tracked `*.go`), so there is no template/asset regeneration to pair with this.

Requested by the coordinator as a retro-package write-up of the review-technique defect, rather than leaving it in mail.


AlexAvilov:
Dual-cleared: opus-xhigh authored, codex SHIP'd at 98ea410. I verified the delta myself (5107804..98ea410, one file), all five checks green, and ran the new step 5's own instruction while holding claimed mail — `agentchute gate --before continue` executes mid-turn, blocks on unread mail, and only warns about the lane's own claimed mail, exactly as the step now describes. codex's finding was real: the first version told a guarded reviewer to run `check`, which the guard denies until turn-end. Merging.

PR #148, opened 2026-08-17
refactor(op): extract the operation seam (M1, WI-1.1..WI-1.8)

AlexAvilov:
M1 — the operation seam. `internal/op` becomes the only way the CLI mutates or
reads pool state; `internal/cli` keeps parsing and rendering; `internal/loop` is
untouched. No wire code, no transport code, no SSH anything.

**Base SHA:** `d2614d5` (`main`, PLAN.md's own merge, #144).
**Scope:** WI-1.1 … WI-1.8. **WI-1.9 (release notes) is integrator-owned and NOT in
this PR** — claude-code lands it here before the merge.

---

## Verification

`sh tools/test.sh` — **PASS** (gofmt + `go vet ./...` + `go test ./...` + `go build ./...`,
all env-stripped):

```
gofmt -w .
go vet ./...
go test ./...
?   	github.com/agentchute/agentchute	[no test files]
ok  	github.com/agentchute/agentchute/internal/cli	11.543s
ok  	github.com/agentchute/agentchute/internal/loop	(cached)
ok  	github.com/agentchute/agentchute/internal/op	(cached)
ok  	github.com/agentchute/agentchute/internal/runner/pty	(cached)
go build ./...
test.sh: PASS
```

Counts, fresh (`-count=1`), 838 test cases run, 837 pass + 0 fail:

| package | passing cases |
|---|---|
| `internal/cli` | 593 |
| `internal/loop` | 174 |
| `internal/op` | 68 (new) |
| `internal/runner/pty` | 2 |

**Pre-existing flake, NOT from this branch.**
`TestSweepStaleRegistrationsUnderLockRecheckSurvivesHeartbeatOnlyRevival`
(`internal/loop/sweep_test.go:360`) failed once in a full-suite run. Hammered
**at this head**: 1 failure in 40. Hammered **at the unmodified base `d2614d5`**
in the primary checkout: 1 failure in 60. `internal/loop` has a zero-byte diff
on this branch, so the flake is inherited, not introduced. This is the **second
recorded sighting** (first: ubuntu CI, 2026-08-12) — flagging it for its own
work item, out of M1's scope.

## `git diff --stat d2614d5..ac0b6eb`

```
 internal/cli/ack.go            |  63 +++---
 internal/cli/check.go          | 278 +++++-------------------
 internal/cli/clean.go          |  51 +----
 internal/cli/gate.go           | 252 ++--------------------
 internal/cli/register.go       | 147 +++----------
 internal/cli/run_test.go       |  16 +-
 internal/cli/send.go           | 119 +++++------
 internal/cli/send_a5_test.go   |   7 +-
 internal/cli/send_b3_test.go   |   7 +-
 internal/cli/serve.go          | 114 +++++-----
 internal/cli/status.go         |  69 +++---
 internal/cli/turn_end.go       |   5 +-
 internal/op/acceptance_test.go | 140 ++++++++++++
 internal/op/ack.go             |  62 ++++++
 internal/op/channel.go         | 204 ++++++++++++++++++
 internal/op/channel_test.go    | 254 ++++++++++++++++++++++
 internal/op/claim.go           | 294 ++++++++++++++++++++++++++
 internal/op/claim_test.go      | 361 +++++++++++++++++++++++++++++++
 internal/op/cleanowed.go       |  87 ++++++++
 internal/op/deps_test.go       |  80 +++++++
 internal/op/errors.go          |  80 +++++++
 internal/op/errors_test.go     | 168 +++++++++++++++
 internal/op/event.go           | 111 ++++++++++
 internal/op/gate.go            | 230 ++++++++++++++++++++
 internal/op/helpers_test.go    | 157 ++++++++++++++
 internal/op/op.go              |  32 +++
 internal/op/pending.go         |  94 +++++++++
 internal/op/readside_test.go   | 468 +++++++++++++++++++++++++++++++++++++++++
 internal/op/register.go        | 287 +++++++++++++++++++++++++
 internal/op/register_test.go   | 235 +++++++++++++++++++++
 internal/op/send.go            | 140 ++++++++++++
 internal/op/send_test.go       | 252 ++++++++++++++++++++++
 internal/op/status.go          | 131 ++++++++++++
 33 files changed, 4167 insertions(+), 828 deletions(-)
```

- **`internal/loop`: zero files, zero lines.** (`git diff --stat d2614d5..ac0b6eb -- internal/loop` is empty.)
- **Existing `*_test.go` touched: exactly the three the plan names**, each limited
  to the named hunk. Each also gains ONE import line for
  `internal/op` — unavoidable, since the pinned replacement text (`op.SendTsMessageWithCommit`,
  `op.NewChannel(...)`) names the package.
  - `send_a5_test.go` — the seam-var retarget at `:229-237` (+ import).
  - `send_b3_test.go` — the seam-var retarget at `:137-146` (+ import).
  - `run_test.go` — `newPollTestRuntime`'s literal at `:495-507` (+ import).
  Full diffs of all three are ~40 lines; nothing else in those files moved.
- Zero new Go dependencies; both `go.mod` files byte-identical.

## Move-set rule (R1/T2) — the grep, run at `d2614d5` over `internal/cli/*_test.go`

Every identifier this merge removes from, or re-signatures in, package `cli`:

| identifier | test hits | classification | disposition |
|---|---|---|---|
| `sendTsMessageWithCommit` | 7 | code (all 7 in `send_a5_test.go` / `send_b3_test.go`) | **named exception rows 1–2** |
| `regTemplate` | 1 | code (`run_test.go:505`) | **named exception row 3** |
| `runnerRuntime.lease` | 1 | code (`b1_convergence_test.go:67`) | **field RETAINED as the adapter** — see deviation 1 |
| `gateStatus` | 12 | code (4 files) | **type alias** `type gateStatus = op.GateResp` |
| `displayConsumed` | 1 | comment (`check_age_owed_test.go:241`) | nothing needed |
| `displayConsumedReadOnly` | 1 | comment (`check_age_owed_test.go:178`) | nothing needed |
| `publishRegistrationOnce` | 1 | comment (`clean_test.go:476`) | nothing needed |
| `lastSweep` | 2 | comment (`b1_convergence_test.go:21,40`) | nothing needed |
| `sweepInterval` | 1 | comment (`b1_convergence_test.go:21`) | nothing needed |
| `hasPendingInboxMail` | 2 | comment (`run_test.go:595,845`) | method kept as a one-line adapter |
| `printReplyRefIfRequired` | 0 | — | subsumed (deviation 4) |
| `isFrontmatterReplyRequired` | 0 | — | moved |
| `evaluateGate` | 0 | — | moved |
| `evaluateGatePhase` | 0 | — | moved (deviation 3) |
| `finishGateClear` | 0 | — | moved |
| `phaseChecksStaleReg` | 0 | — | moved |
| `archiveAllClaimed` | 0 | — | moved |
| `readRegistrations` | 0 | — | moved |
| `registrationLiveElsewhere` | 0 | — | moved |

Signature-preserved on purpose, so they stay OFF the list (population 2 is
empty for this merge): `performRegister` (17 hits), `printStatus` (4),
`refuseLiveRunnerCollision` (4), `StaleRegThreshold` (5, now `= op.StaleRegThreshold`),
`cmdCleanOwed`, `heartbeatTemplate`.

## Dependency direction (B1/B2), enforced

`internal/op/deps_test.go` parses the package's own files with `go/parser`
(stdlib only) and fails if `internal/op` imports anything outside
{stdlib, `internal/loop`}, and if `internal/loop` ever imports `internal/op`.

---

## Deviations from PLAN.md — five, each deliberate; please gate these first

**1. `runnerRuntime.lease` is RETAINED (not collapsed).** WI-1.7(d) says one
`*op.Channel` replaces three fields. `regTemplate` and `lastSweep` do collapse.
`lease` cannot: **`b1_convergence_test.go:67` calls `loop.ReleaseLease(rt.lease)`** —
a fifth lease call site the plan's R1c/T5 enumeration (which lists only
`run_test.go:130,140,159,170`) missed, and WI-1.7(f) requires that file green
**unmodified**. The move-set rule's own branch covers it: leave a thin adapter
rather than widen a closed exception table. The channel ADOPTS the same lease
(its documented adoption arm), so there is one lease with two names; `runWrapper`'s
shutdown release stays byte-unchanged.

**2. `op.Register` takes `now` as a fourth, NON-WIRE argument.** Design fixes the op
signature at `(cfg, ctx, Request)`, with `now` hub-minted. That is unsatisfiable for
register while `performRegister`'s pinned signature keeps its `now`:
`register_test.go:369-377` asserts the written `LastSeen` **equals the injected value**,
and `register_test.go:52-58` injects **different** values from concurrent goroutines —
so a package-level clock seam is out too. `now` therefore arrives the same way
`Context` does: an explicit argument that never rides a request struct. The
`RegisterReq` field list is EXACTLY as pinned (no `HostProvided`, no `Now`). Every
other op mints its own clock as designed. Cheapest alternative I could find; flagging
it as the one shape change.

**3. `evaluateGatePhase` MOVED to `internal/op` (WI-1.4(c) lists it as staying).** It is
the blocking predicate, not a renderer: `evaluateGate` calls it, and `internal/op`
cannot import `internal/cli`. Leaving a copy behind would fork the gate's
blocking decision into two sources — the exact thing `evaluateGate`'s doc comment
exists to prevent. Zero test hits, so no alias is owed. `cli` keeps `isValidGatePhase`
(pure string validation, no state) and re-exports the five phase constants.

**4. `displayConsumed`, `displayConsumedReadOnly` and `printReplyRefIfRequired` are
subsumed by one renderer** (`renderClaimedMessage`), where WI-1.3(c) has the latter two
staying. With the seam there is exactly one render path: the mutate-vs-don't split is
now the op's (`ClaimReq.NoArchive`), and the reply ref is a MessageEvent field because a
remote client never sees the filename to parse. Keeping CLI copies would be dead code
plus a second source of truth for the ref. `printConsumedBody` and `sanitizeControlBytes`
stay exactly as the plan says.

**5. Owed discharge is per-call-site, not globally suppressed under `NoArchive`.**
WI-1.3(c) says the ClearOwed half "is suppressed under `NoArchive`". Applied globally
that CHANGES behavior: today the redelivered-residue path calls the mutating
`displayConsumed` (`check.go:192`) **regardless** of `--no-archive`; only the fresh-message
path uses the read-only variant. Zero-behavior-change (constraint 1) wins, so `op.Claim`
discharges on the residue path always and on the fresh path only when it claims. Pinned by
`TestClaimResidueDischargesEvenUnderNoArchive` + `TestClaimDischargesOwedOnlyWhenItMutates`.

### One contract the plan/design did not pin, which I had to

`op.Send`'s **third return shape**: `SendResp` is fixed at
`{Filename, Ref, Committed, DurabilityNote}`, and `DurabilityNote` is spoken for
(linked-but-sync-failed). The asker-side owed-bookkeeping failure is a DIFFERENT
partial success with its own shipped warning text, and no sentinel is available
(the set is closed at eight). So: `op.Send` returns a **populated** `SendResp`
**beside** a non-nil error, which is RecordOwed's own error unwrapped;
`resp.Filename != ""` is what tells "delivered, bookkeeping failed" from "not
delivered". Documented on the func and pinned by
`TestSendReportsOwedBookkeepingFailureBesideACommittedResponse`. M3 will need to
frame this; worth a reviewer's eye now.

### Coverage gaps I did not paper over

- **C4 fresh-suffix collision retry** (WI-1.2(e)) has no seam-level lever: it lives
  inside `loop.DeliverUnderRecipientLock`, forced via `loop`'s unexported
  `rand128hex`, and `internal/loop` is frozen in M1. It stays covered by
  `internal/loop/send_delivery_test.go` and the untouched CLI send tests. Not
  faked at seam level.
- **`Sweep:true` whose sweep FAILS** (WI-1.6(e)) is not tested: the only error path in
  `SweepStaleRegistrations` is a failed `ReadDir` of the agents dir, and the
  registration write in the same op needs that dir writable first. The
  success/skip arms and the warning FORMAT are covered; the failure arm needs a
  `loop` seam that M1 forbids.
- The byte-for-byte stdout/stderr golden (WI-1.3(e)) is asserted as **event order
  plus verbatim `Msg` text** in `internal/op`, per WI-1.8(c)'s "NEW
  `internal/op/*_test.go` additions only". The rendered bytes stay pinned by the
  existing, unmodified `check_a6_test.go` / `check_age_owed_test.go` /
  `consume_boundary_test.go` / `n3_sanitize_test.go`.

## Item-by-item

- **WI-1.1** `op.go` / `event.go` / `errors.go`: `Context{ActorID}`, the four-arm
  event union with constructors that cannot violate it, the eight sentinels (four
  re-exports, four new, both stale arms wrapping `*loop.ErrRecipientStale`), and
  `CodeFor` with its `E_HUB_IO` default. The sentinel set is held closed by a test
  that reads the package's own source, so a ninth without a code arm fails.
- **WI-1.2** `op.Send` + exported `op.SendTsMessageWithCommit`. `op.SendPreflight` is
  exported so `cmdSend` keeps its A5 fast-fail BEFORE stdin; `op.Send` re-runs it as the
  authoritative check (on the wire there is no earlier point — DESIGN §4.5.1).
- **WI-1.3** `op.Claim`, streaming. The three stdout lines are `info` notes at their
  production points; every stderr warning site is a `warn` note with no level prefix.
- **WI-1.4** `op.Status` / `op.Gate` / `op.Pending` / `op.CleanOwed`. Status never
  truncates (both framing budgets are the wire producer's) and streams its lenient-read
  warnings so they land before the table. `gateStatus` survives as an alias.
- **WI-1.5** `op.Ack`: per-item `AckItemEvent` as each commits, then the finish gate.
- **WI-1.6** `op.Register` + `RegistrationView`/`AnnounceView`; `performRegister` stays
  in `cli` as the host-resolving adapter, all thirteen call sites unchanged.
- **WI-1.7** `op.NewChannel` + `Tick`. Only `ErrFenced`/`ErrOrder` are hard errors;
  the three step failures are `Warnings` carrying the runner's exact log strings.
  Local serve startup order (lease → child → register) is untouched.
- **WI-1.8** Both-`Context`-constructor matrix over every actor-scoped op (identical
  disk effects, asserted non-vacuously) + the dependency-direction test.

## Authorization

Not merging, not tagging, no release step. Gate asks going to **opus-xhigh**
(primary) and **codex** (mandatory second) at head `ac0b6eb`.


AlexAvilov:
Dual-gated at 46730c5: **codex SHIP** (`d588328..46730c5`, all prior findings closed) and **opus-xhigh SHIP** (`5ffa196`, nothing outstanding). CI green on all five checks at this exact head, verified via the REST check-runs endpoint. `internal/loop` is a zero-byte diff from `d2614d5` at every head from first to last, and the closed test-exception table stayed at three files throughout.

Eight deltas: five codex findings (the nil lease that left `serve.claim` on disk and made the next `serve` refuse to start; vendor presence-versus-emptiness; the tick order check nested under the lease branch; `op.Send`'s unrepresentable-on-the-wire return shape; a stale `send.go` comment describing the shape that change deleted), three opus-xhigh items (the guard latch arming after a body read rather than from the listing; a twin canonical-id predicate; a package-scope hole in the constructor guard), one of mine (a data race the new end-to-end test made reachable, which `release.yaml`'s own `-race` job would have failed *after* the tag), and two stale published numbers in the release notes.

Merging, then tagging v1.6.0.

PR #159, opened 2026-08-17
feat: add SSH hub wire session

AlexAvilov:
## Summary

- add the v1 NDJSON + raw-trailer hub codec, full frame/code vocabulary, handshake, status two-budget encoder, fuzz corpus, and streaming boundary tests
- add `agentchute hub session` with direct pinned-pool config, J1 pool identity validation, every-op dispatch, channel ordering, deadlines, and lease release on every exit
- add boot-reference lease proof on Linux/Darwin, invocation-scoped send delivery injection, and executable L1-L4/W1-W6 conformance vectors

### Shared W session factory

`internal/spectest/wire.go` defines the M3/M6 transport boundary:

```go
type SessionFactory interface {
    Open(context.Context, string) (Session, error)
    State() StateProbe
}

type Session interface {
    io.ReadWriteCloser
    ForceDisconnect() error
    Wait(context.Context) error
}
```

`StateProbe` exposes only the fixture/state operations the six shared W assertions require. M3 drives it with `net.Pipe`; M6 can supply the SSH transport without duplicating assertions.

### No-mutable-send-state check

`TestSendHubPathHasNoMutablePackageState` parses the root-module `internal/op` call graph starting at `Send`. It permits only the fixed error sentinels and fails if any other package variable—exported or unexported—becomes reachable. Production `Send` passes `loop.SendTsMessageWithCommit` into the invocation-scoped `sendWithDelivery` helper; CLI-only tests inject through `cmdSendWithOp`.

## Scope proof

- base: `7c8c4451f555cf24d384d3363528368e8eb824af`
- head: `e509239`
- no changes to `AGENTCHUTE.md`, PLAN, DESIGN, either `go.mod`, or either `go.sum`
- `internal/loop` production delta is exactly `lease.go` plus `bootref_{linux,darwin,other}.go`; `lease_test.go` contains the seven deterministic WI-3.4 rows and mixed-version-key assertion
- `tools/test.sh` has exactly the planned env-stripped conformance block

## Verification

- `sh tools/test.sh`
- env-stripped `go test -race ./... -count=1 -timeout=300s`
- Linux amd64 compile: `internal/loop`, `internal/cli`
- Windows amd64 compile: `internal/loop` (verifies the third boot-ref build-tag arm)
- Darwin tests executed locally

### DIFF
--- a/conformance/CONFORMANCE.md
+++ b/conformance/CONFORMANCE.md
@@ -14,6 +14,14 @@ Four deliverables in one package:
 
 Pure stdlib, no dependencies.
 
+## Lease and hub-wire vectors
+
+`vectors/lease.json` and `vectors/wire.json` extend the executable contract
+with lease/fencing and session-lifecycle cases. Their reference drivers live in
+the root module under `internal/spectest`, because the standalone conformance
+module cannot import the reference implementation's `internal/*` packages.
+Other bindings should drive the same JSON through their own harness.
+
 ## Run
 
 ```sh
--- a/conformance/vectors/lease.json
+++ b/conformance/vectors/lease.json
@@ -1,0 +1,9 @@
+{
+  "schema": "agentchute-lease-vectors-v1",
+  "vectors": [
+    {"id":"L1","kind":"fresh_acquire_fails_closed","name":"a second acquire while fresh fails closed"},
+    {"id":"L2","kind":"fenced_holder_writes_nothing","name":"a fenced holder cannot heartbeat or mint"},
+    {"id":"L3","kind":"release_token_checked","name":"release never deletes a successor claim"},
+    {"id":"L4","kind":"stale_reclaim_fences_old","name":"stale pid-dead reclaim succeeds and kills the old token"}
+  ]
+}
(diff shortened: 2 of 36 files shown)


AlexAvilov:
Dual-gated at 7fb6395: **grok SHIP** on the codec and round-trips, **opus-xhigh SHIP** on both delta hops. All five checks green; I independently ran `go test -count=1 -race ./...` clean on all six packages and the conformance module green in its own directory.

The one blocking finding was opus-xhigh's, and it is the kind that only shows up one frame away from where everyone is looking: `register-ok` carried `reg.host` and `reg.working_repos` inline and unbudgeted, and encoded **after** `op.Register` had already written the row. A 71,680-byte host produced `control frame exceeds 64 KiB` on `register-ok` while `status-ok` correctly truncated the same input — so an affected lane would have been simultaneously invisible in `status` and unable to re-register, including at turn-end step 0. Fixed by a transport-supplied precommit validator that encodes the exact wire response under the agent lock, after the merge and before any filesystem write, so an unencodable response leaves no row behind. `Register`'s signature is unchanged and the local path passes nil.

Also landed: an invariant comment recording that `StatusOK.Truncated` must stay present when false, because `EncodeStatus` measures candidates with `false` — one byte longer than `true` — so adding `omitempty` in a later tidy would silently flip the 64 KiB budget from conservative to optimistic.

Merging.

Issue #152, opened 2026-08-17
M3 hand-off: the seven items M1 discovered and deliberately deferred


AlexAvilov:
Everything M1 discovered but deliberately did not fix, in one place, so codex inherits a list rather than archaeology. M1's implementer (opus-high) declared each of these rather than papering over it; none is a defect in M1's shipped code.

## Wire-shape amendments (decided, not open)

**1. `op.Send` — `owed_note`.** `SendResp` gains `OwedNote string`; the `send-ok` frame gains `owed_note`; the operation error is nil once delivery commits. **Authorized by claude-code, implemented in M1.** Rationale: §§4.4/4.5.3 terminate a remote send with either `send-ok` or `error`, and the old shape (populated response beside a non-nil error) is unrepresentable — an `error` frame loses the committed response and drives spool/failure handling, a duplicate-send hazard on an already-committed operation, while the fixed `send-ok` shape had nowhere to carry the warning. Found by codex; opus-xhigh's contrary ruling rested on "M3 frames it anyway", which is false.

**2. The error path must carry "claimed mail is held".** After M1's guard-latch fix, the unreadable-residue signal lives in `ClaimSummary` beside a non-nil error. Locally that is fine — Go returns both. On the wire it is not: the terminal error frame carries no summary, so a **remote** client cannot arm its latch over residue it never received a `MessageEvent` for. That is the same unrepresentability class as item 1, and it would leave a remote lane holding claimed mail with the guard down — the exact state the latch exists to cover.

**Ruling (claude-code): pin the amendment now, implement it in M3/M4.** The frame vocabulary gains an optional boolean on the error path (codex's suggested `claimed_held`); M3 frames it, M4 arms from it. No M1 code change — there is no wire yet to carry it.

Two constraints on whoever implements it:
- **A `note` frame is not sufficient.** Notes already exist in the vocabulary and would need no amendment, but arming a latch must never depend on parsing display text. This needs a machine-readable flag.
- **The exact field name and placement are codex's to propose** in the M3 gate ask, since M3 owns the codec. Pinning happens in DESIGN + the M2 spec copy before M3 lands.

**Why pin now rather than at M4:** PR #146 (the M2 spec) is SHIP'd but parked behind the v1.6.0 tag, so it has not shipped and amending it is nearly free. Deferring means M2 releases a frame vocabulary that cannot express a required behavior, and the amendment then lands *after* a release describes the old one — the C6 problem in reverse.

## Carry-items that must not evaporate

[...]

PR #160, opened 2026-08-17
feat: add SSH hub client transport

AlexAvilov:
## Summary

- add strict `ssh://` discovery, canonical hub identity, joined config, and local shadow state
- add one-shot SSH transport, client sessions, error mapping, mux fallback, negative cache, and send ambiguity spool
- route remote CLI verbs through the hub while preserving hub-side registration, announce, sweep, status, and local guard semantics
- add names-map identity resolution, remote identity/doctor surfaces, launcher forwarding, and W1/W2/W6 client conformance

## Implementation note

The pointer discovery arm branches on `ssh://` immediately after `DiscoverPointer`, before its local `AGENTCHUTE.md` and `.agentchute/loop` checks. This is deliberate: preserving the locator in `ResolvePointerTarget` alone would only replace the old mangled-path failure with a different local-path refusal. `TestDiscoverRemotePointerBranchesBeforeLocalRepoChecks` pins the complete path.

The review/CI delta joins every wire-test hub session before `t.TempDir` cleanup, makes W2 disconnect on the terminal read only after the hub commit is observable, and keeps remote wrapper children on the canonical SSH locator without exporting the local shadow loop. The prior W3 cleanup writer was the deferred `ReleaseLease`/`withAgentLock` path from a session that outlived its test.

The final review delta replaces the isolated `claimed_held` helper test with a W6 integration assertion that drives the production remote `cmdCheck` over a fake SSH wire peer and verifies the real local guard latch. Removing the `check.go` claimed-residue branch now fails the test.

## Verification

- `tools/test.sh`
- env-stripped `go test -race ./...`
- env-stripped `cd conformance && go test -race ./...`
- Linux non-root `go test -race ./...`
- Linux `TestWireVectors/W3/reclaim_during_send -count=20 -race`
- Linux `TestWireClientVectors/W2/client -count=20 -race`
- `TestW6ClientRemoteCheckClaimedHeld -count=20 -race`

Base: `f90a96866df93d43a51c188f5af0bd0b7c9bc9d4`
Head: `86269248f7127fa199db3994dfafb904d4ac64f7`

### DIFF
--- a/internal/cli/ack.go
+++ b/internal/cli/ack.go
@@ -88,7 +88,7 @@ func cmdAck(args []string) error {
 		return err
 	}
 
-	agentID, err = resolveAgentID(agentID)
+	agentID, err = resolveAgentID(agentID, cfg)
 	if err != nil {
 		return err
 	}
@@ -155,12 +155,23 @@ func cmdAck(args []string) error {
 // empty, non-nil slice).
 func ackClaimed(cfg *loop.Config, agentID string) ([]ackItem, op.AckSummary, error) {
 	acked := make([]ackItem, 0)
-	sum, err := op.Ack(cfg, op.Context{ActorID: agentID}, op.AckReq{}, func(ev op.Event) error {
+	emit := func(ev op.Event) error {
 		if ev.Ack != nil {
 			acked = append(acked, ackItem{Filename: ev.Ack.Filename, ArchivePath: ev.Ack.ArchivePath})
 		}
 		return nil
-	})
+	}
+	var sum op.AckSummary
+	var err error
+	if cfg.Remote != nil {
+		session, openErr := openRemoteOneShot(cfg, agentID)
+		if openErr != nil {
+			return nil, sum, openErr
+		}
+		sum, err = session.Ack(emit)
+	} else {
+		sum, err = op.Ack(cfg, op.Context{ActorID: agentID}, op.AckReq{}, emit)
+	}
 	if err != nil {
 		return nil, sum, err
 	}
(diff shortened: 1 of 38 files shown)


AlexAvilov:
Dual-gated at 8626924: **grok SHIP**, **opus-xhigh SHIP** with its HIGH closed. All five checks green; I independently ran `go test -count=1 -race ./...` clean on all seven packages and W2+W3 at `-count=30 -race`.

Three real defects surfaced in this merge, and none was in the feature being built:

1. **A conformance vector that could pass while asserting the opposite of its own claim.** W2's write-side disconnect poll raced the hub's `send-ok`, so Linux sometimes observed success where the vector asserts `E_SEND_UNKNOWN` — in the never-auto-replay area, where a wrong answer means a duplicate message. Now deterministic: the terminal read waits for observable hub commit, force-disconnects, then reads the closed transport.
2. **A session outliving its own cancel.** W3's temp dir would not delete on ubuntu because the harness closed without joining, and the deferred `ReleaseLease` re-entered `withAgentLock` and *recreated* the lock file during `t.TempDir` removal — cleanup itself was the writer. Wire sessions are now joined before cleanup. Third instance this program of work outliving its scope; see #152 item 8.
3. **A remote child inheriting the wrong discovery context** (opus-xhigh, HIGH): `runnerChildEnv` now strips inherited `AGENTCHUTE_CONTROL_REPO`/`AGENTCHUTE_LOOP_DIR` and re-adds the hub URL alone on the remote arm, with no shadow loop-dir — §6.8 rule 5.

Also closed a W6-client assertion gap grok found: the client half now drives production `cmdCheck` through a fake SSH peer and asserts the real guard latch, mutation-probed.

Merging.

PR #161, opened 2026-08-17
feat: add SSH hub lifecycle and join UX

AlexAvilov:
## Status

Draft M5 implementation for `proposal/ssh-hub/PLAN.md` WI-5.1 through WI-5.7.

Implemented in the first slice:

- persistent remote serve channel with hello, lease, register-before-child, tick, and token-checked release
- remote child environment contract and last-tick pending cues
- fence-on-channel-drop with SIGTERM before exit/relaunch
- default-on remote relaunch, opt-out behavior, capped jittered backoff, and one `E_FENCED` retry
- in-process lifecycle, cue, channel-loss, and relaunch tests

Remaining on this branch:

- `hub join`, key lifecycle, and same-hub migration
- `hub authorize` and wipe preservation in the same merge slice
- error catalog, doctor coverage, and final M5 integration rows

## Verification

- `tools/test.sh`
- env-stripped `go test -race ./...`

## Authorized scope widening

- WI-5.3 migration review fix: `internal/loop/runner.go` was outside M5's literal MODIFY list. The integrator authorized the nine-line widening: add `ErrRunnerStateAgentMismatch` and wrap the existing mismatch error with `%w`. This was necessary to distinguish an absent runner file from corrupt JSON and an agent-id mismatch without brittle error-string matching or duplicating the runner-state loader.

## Compatibility note

- The runner-state agent-mismatch error's rendered text now ends with the sentinel's own string because of the `%w` wrap. No test asserted the former exact text; any documentation quoting that error verbatim should be refreshed.


AlexAvilov:
Fully gated at `4f01561`, five pinned verdicts across four surfaces:

- migration fail-closed `83d2b3c..1bf390c` — **opus-xhigh SHIP**
- WI-5.4 key/authorization lifecycle `1bf390c..4bc50e6` — **opus-xhigh SHIP**
- WI-5.5 `state/pool.id` wipe preservation `1bf390c..4bc50e6` — **opus-xhigh SHIP**
- WI-5.6 §7 persona walk, then `ad398bc..4f01561` — **grok FIX ×2, then SHIP**
- WI-5.7 session-clock concurrency `e703f7e..ad398bc` — **opus-xhigh SHIP**

All five checks green; I ran `go test -count=1 -race ./...` clean on all seven packages at each pushed head.

Three findings worth recording, because none was in the feature being built:

1. **A migration that failed open.** `refuseLiveHubMigration` treated three distinct `LoadRunnerState` errors — absent, corrupt, agent-id mismatch — as "no runner is running", so a join could proceed while unable to prove a lane was stopped, stranding that lane's latch and spool in a renamed directory. §7.2 says ambiguity fails closed; this was the one place it did not. Now only `fs.ErrNotExist` continues.
2. **A restart instruction that would not restart anything.** The `E_FENCED` text told the operator to run `ac serve <id>`, which is not a working command; a lane restarts with `ac --as <id> serve <wrapper>`. Caught by the persona-walk lane reading it as something to *type* rather than as a sentence to check for accuracy.
3. **A process-global clock inside the session-serving path.** `hubSessionNow` reintroduced the exact shape WI-3.6 had just removed from `op.Send`, in the file where the multi-session hub process actually runs. Now an invocation-scoped `hubSessionOptions.now`, with the one-shot CLI seams left as globals — which is the correct distinction, now demonstrated twice.

Merging.

PR #162, opened 2026-08-18
M6: real SSH hub integration matrix + the fd-race, migration data-loss, and hub write-lock fixes it surfaced

### LINKED ISSUE #165: DATA LOSS: hub join's migration recovery branch deletes the old hub dir without copying it, exit 0
**`hub join` can delete a hub's authoritative state directory — including the authorized private key — without copying it, exit 0, and print `joined:`.** Reproduced on `origin/main` at `5e8dfea`. No fixture state involved.

## Mechanism

`internal/cli/hub_migrate.go:111-121`:

```go
if _, err := os.Stat(newDir); err == nil {
    // A committed new directory means this is post-rename recovery: only the
    // pointer and old-directory cleanup remain.
    writeHubJoinPointer(...); hubMigrationReapMux(...); os.RemoveAll(oldDir); return
}
```

The copy (`copyHubMigrationTree`) is in the `else` path. The branch keys on **nothing but the destination directory existing**, so it cannot distinguish *"I crashed after my own rename"* from *"something else created this directory"*. In the second case it writes the pointer, reaps the mux, and **deletes the old hub dir that was never copied.**

## The trigger is a FAILED join, not a deliberate one

`prepareHubKey` runs at `hub_join.go:143`, **before** `finishHubJoinKey` contacts the hub at `:157`. So the new hub dir and its key material exist *before the hub is reached at all*. Any failure at or after the hello — network, refused authorize, non-writable pool, pool mismatch, Ctrl-C — leaves `newDir` on disk with a key and **no `config.json`**.

The operator sees an error and retries. **The retry destroys the old directory.** "Retry after a failed join" is the first thing anyone does.

It also compounds: `hub_join.go:103-104` runs `migrateHubJoinState` **then** `runHubJoin`, so if the retry fails again for the same reason, the old directory is already gone and there is no path back.

## The bitter detail

In the missing-`config.json` sub-case, `findHubMigrationCandidate` takes the `!currentExists` arm and probes with the **old** key (`hub_migrate.go:74`). It returns the old hub id *because that probe succeeded* — the code has just demonstrated that key is the working credential, and then deletes it.

## Observed

```
PRECONDITION: newDir .../hub/0d1399d9271d exists before the alias join
pointer replaced: ssh://alex@hub.example/... -> ssh://alex@hub-alias.example/...
joined: codex-tiny (local name: codex) @ ssh://alex@hub-alias.example/... (agentchute-hub v1)
old hub dir removed = true
old active pub == new active pub = false
DATA LOSS CONFIRMED: authorized key "codex-tiny_ed25519.v1" no longer exists anywhere
```

Exit 0. Nothing printed says a directory was deleted or a copy skipped.

## Required fix — BOTH parts, not either

**(a) The recovery branch needs positive evidence of its own rename**, not a directory's existence: a marker the rename consumes, or provenance in the copied `config.json`. Note that simply falling through is not safe — `os.Rename(partial, newDir)` onto a non-empty directory fails `ENOTEMPTY`, yielding a confusing errno rather than safety.

**(b) The irreversible step gets its own precondition**: before `os.RemoveAll(oldDir)`, verify the new dir actually contains the old dir's key material. No schema change, covers directories predating (a), and blocks both sub-cases.

Two checks because they fail differently: (a) checks **intent**; (b) checks **the thing about to be destroyed is not unique**. A fix implementing only (a) is not sufficient.

## Why it survived

`migrateHubJoinState` and `findHubMigrationCandidate` are referenced by **no test file in the tree by name**. The single migration test, `TestHubJoinSameHubMigrationPreservesStateAndKey`, drives only the copy path — `newDir` never pre-exists — and asserts exactly the invariant the recovery branch violates (`newPub == oldPub`, latch preserved). The branch is not under-tested; it is **untested**, while its sibling carries the assertion that would have caught it.

The regression row is therefore that same test with `newDir` created first. A working probe exists and can be handed over verbatim.

Found during M6 by opus-high (post-migration key replacement), mechanism proposed by claude-code, reproduced and extended on `origin/main` by opus-xhigh.

### LINKED ISSUE #169: E_CHANNEL_LOST discards its Cause, so the one unclassified ssh error arm is the least diagnosable
`hubclient/ssh.go:320` is the final fallback arm of the ssh error classifier: reached after permission-denied, exit-127, hello-timeout and connect-failure have all been ruled out. It returns `E_CHANNEL_LOST` with the operator-facing text:

```
hub: channel to the hub was lost
```

The error carries a `Cause`, and the message does not print it. So the one arm that by definition means **"an ssh failure we did not classify"** is also the one that discards the only information about what it was.

## Cost, measured

An intermittent CI failure (~15% of ubuntu runs, `TestSSHDChildEnvSendAndSupervisedRelaunchDefault`) produced exactly this string. Two rounds of diagnosis were spent reasoning from it, including two fix attempts aimed at a mechanism that turned out to be normal behaviour, because the sentence is compatible with every hypothesis. The underlying `Cause` was present in the process the whole time.

A production operator hitting this arm gets the same sentence with the same lack of recourse, and unlike us they cannot go read the ssh log.

## What makes it worse than a generic wrapped error

The stdout in the failing case shows the operation **succeeded** before this fired — the `check` reached the hub, ran, claimed a message and printed it, and only then exited non-zero. So the fallback arm can fire on teardown of a successful operation, and the message gives the operator no way to tell "nothing happened" from "everything happened and then the connection dropped". Those imply opposite next actions: re-run, versus do not re-run because the mail is already claimed.

## Suggested fix

Include the `Cause` in the rendered message for this arm specifically. The classified arms have meaningful text and do not need it; the unclassified one is defined by not knowing, so it should say what it saw:

```
hub: channel to the hub was lost: <cause>
```

If there is a reason the cause is unsafe to surface verbatim (it can carry remote paths), then it should at minimum be logged under a verbose flag, and the operator-facing text should say where to look.

## Related

Part of the pattern in #167 — a result inspected in a way that cannot distinguish its cases. This is the fifth instance found in M6.

AlexAvilov:
M6: the real-sshd integration matrix, plus the production defects it and its reviews surfaced — four rounds deep.

## What this PR contains

**The matrix** (original subject): ~15 integration test files driving a real sshd on both platforms — join/authorize/doctor, wire conformance, multiplexing, key rotation (A2), fence relaunch (A3), join idempotence (A4), lease lifecycle, migration recovery, and the harness for all of it. Now runs under `-race` in CI.

**Production fixes found by the matrix + the 3-machine campaign + four review rounds:**
- **red #3**: the hub client's fd race — `cmd.StdoutPipe` + goroutine-`Wait` closed the read end with bytes undrained; the transport now owns its pipes (the same fix was then needed in the test harness itself, where `-race` exposed the identical pattern).
- **#165**: migration data loss — the recovery branch keyed on directory existence; now a provenance marker plus byte-level content verification gate the irreversible delete.
- **The migration freeze**: verify-then-delete was a TOCTOU; the old tree is renamed to `<newDir>.migrating` before verification, so what is deleted is exactly what was verified, and a verify-first sweep recovers the crash states.
- **The hub write-lock contract (§13.9a)**: a running lane's state lives *inside* the migrated tree, and held descriptors follow inodes — so `serve` holds the hub SHARED for its lifetime, every remote one-shot takes SHARED at the single `discoverConfig` seam for the command's lifetime, and a migration takes EXCLUSIVE on both hub ids and refuses while any holder lives. The contract is normative in AGENTCHUTE.md §13.9a, on `ShadowLoopDir`, and operator-facing in docs/hub.md. Residual (a writer violating the contract) is documented and tracked in #179.
- **E_CHANNEL_LOST diagnosability** (#169) and six operator-message fixes from the campaign (doctor exit contract, migration half-finished messaging, join UX, older-binary warning, README corrections).

**Review record**: grok SHIP; opus-xhigh seven passes (design + deltas), every property mutation-pinned by a row proven to be the only one that catches its mutation, verified under CI-like environments; codex backstop found the verify→delete race and the manual-send spool writer, both fixed, final verdict CLOSED. Five green-for-the-wrong-reason test rows were caught by running mutations rather than reading rows — now a standing rule.

**Deliberately out**: pinning-bypass detection (spec ready, next PR), the structural write barrier (#179), harness migration reachability (#180, root cause #164), loop flake (#175).

Closes #165. Closes #169.


AlexAvilov:
## Deep review — opus-xhigh — **FIX** (two one-line fixes; nothing architectural)

Envelope verified: `git diff 5e8dfea3e0a67a292cf682c22237ca7177091b22..04804d6d5461e0a73c03b754f1a44231300db7bd` = 38 files, +4847/−107, exactly the claimed scope. No surplus files.

**The two headline fixes are correct and genuinely pinned.** I re-derived both by mutation rather than trusting the PR text:

- **red #3 (458d7c8)** — restoring `cmd.StdoutPipe()` in `startProcessTransport` fails `TestProcessTransportDeliversOutputWrittenBeforeExit` at *iteration 0* with `read |0: file already closed`, exactly as claimed. The row drives the production constructor, not a copy. Every error path in `startProcessTransport` closes all four descriptors (second-`os.Pipe` failure closes both ends of the first; failed `Start` closes all four; success leaves only the two the transport owns). `Close` is race-clean: the second caller blocks on `closeDone`, which is closed after `p.err` is written, so the read of `p.err` is ordered. EOF semantics are now correct — the child's write end is the last one open, so a late reader drains the buffer and then sees a real EOF.
- **#165 (60f2fd1)** — both guards are independently load-bearing. Deleting the `checkHubMigrationProvenance` call fails only `TestHubJoinRecoveryBranchWithUnmarkedCompleteCopyPreservesOldKey`; deleting the `verifyHubMigrationCopy` call fails only `TestHubJoinRecoveryBranchRefusesWhenTheCopyIsIncomplete`. Neither substitutes for the other. `crashAfterHubMigrationRename` driving the *production* first pass (rather than hand-writing the marker) is the right call — a hand-written fixture would let the production write be deleted with every row still green.

Crash grid for the migration, walked state by state: crash before the marker write, after it, after the rename, after the pointer, after `RemoveAll` — every one of them either re-runs the copy from scratch or re-enters `finishHubMigration` behind the byte-comparison. **No sequence reaches `RemoveAll(oldDir)` with unproven contents.** One state does leave residue that bites later — F2 below.

Suites run at the pinned head: `go test -race ./internal/...` green; `sh tools/sshd-test.sh` green (93s), also green at `-count=2` (194s) and under `-race` (100s) — *after* F1.

---

### F1 — FIX — the sshd suite does not run on a developer machine with a populated ssh-agent

`integration/sshd/harness_test.go:316-323` builds the client `ssh` shim so it injects `-i <adminKey>` when the caller passed no `-i`, but never `-o IdentitiesOnly=yes`. `commandEnv` (harness_test.go:~840) strips only `HOME` and `AGENTCHUTE_*`, so `SSH_AUTH_SOCK` is inherited. The harness `sshd_config` sets `MaxAuthTries 2` (harness_test.go:348).

Every path that shells out to a bare `ssh` with no `-i` therefore offers the developer's *agent* keys first and exhausts `MaxAuthTries` before the admin key is ever tried. The production path that does this is `runHubJoinAutoAuthorize` (`internal/cli/hub_join.go:402`), which deliberately uses the operator's own SSH access.

Measured on both sides, macOS 27 / OpenSSH 10.3, 10 keys in the agent:

- `sh tools/sshd-test.sh` → **11 failures**: `TestSSHDJoinAuthorizeVerifyHappyPath`, `…JoinIsIdempotentAndRotationReplacesTheLine`, `…KeyRotationChangesMuxIdentityAndReauthenticates`, `…RepointTakesEffectAfterLiveMasterIsReaped`, `…RotationConvergesFromEitherRemoteBoundary`, `…RemoteStatusRowsAnd64KiBPrefix`, `…LauncherPathsPreserveRemoteness`, `…FencedLaneRelaunchesExactlyOnce`, `…FencedLaneStopsWhenARivalHoldsTheLease`, `…ChildEnvSendAndSupervisedRelaunchDefault`, `…ChildIsNotRelaunchedWhenOptedOut`.
- Same command with `SSH_AUTH_SOCK` unset → `ok  github.com/agentchute/agentchute/integration/sshd  93.231s`.

The sshd log confirms the mechanism directly: two `Failed publickey` lines carrying the first two fingerprints from my `ssh-add -l`, then `maximum authentication attempts exceeded`. The split is diagnostic — every row that goes through `BuildSSHInvocation` passed, because that function already sets `IdentitiesOnly=yes`; only the bare-`ssh` rows failed.

Why this matters more than a local annoyance: CI is green solely because runners have an empty agent, and the failure *presents as a product defect* — `authorizing via your own SSH access… not available` — so the first human who runs the milestone's headline artifact on their own laptop will debug the product instead of the harness.

Fix: add `-o IdentitiesOnly=yes` to the wrapper's injected options (or `unset SSH_AUTH_SOCK` / `-o IdentityAgent=none` in the script). One line at harness_test.go:~323.

### F2 — FIX — `verifyHubMigrationCopy` does not exempt the migration marker, so a stale `.migrated-from` wedges the next migration

`hub_migrate.go:228` exempts `config.json` "because the migration rewrites its URL by design". The marker is rewritten by design too — `hub_migrate.go:140` overwrites the copied one with the new `oldHubID` — but it is not exempt, so the byte comparison sees `"A\n"` in the old dir against `"B\n"` in the new one and calls it uncopied.

Reachable state: a crash, or an error from the `fsyncHubDir` at `hub_migrate.go:207`, between `os.RemoveAll(oldDir)` and the `os.Remove(newDir/.migrated-from)` at `hub_migrate.go:210` leaves a stale marker in the *live* hub dir. `findHubMigrationCandidate` then finds no candidate (the old dir is gone), so nothing ever cleans it up.

Confirmed with a probe row in `internal/cli` — seed a joined hub, write a stale marker into it, run `hub join <new url>`:

```
hub join: refusing to delete .../hub/d85b3e003cb6: .../hub/d85b3e003cb6/.migrated-from
was not copied to .../hub/0d1399d9271d/.migrated-from. Nothing was changed
```

It is fail-safe — nothing is deleted — but the operator is stuck holding a message that names a file as uncopied when the copy is in fact complete, and nothing in the text points at the actual remedy (delete the stale marker). Fix: put `hubMigrationMarker` in the same skip clause as `config.json` at hub_migrate.go:228.

### F3 — recommend — the sshd suite sits outside both the vet and the `-race` run that CI performs

`.github/workflows/ci.yaml` runs `go vet ./...` and `go test -race ./...` without `-tags sshd_integration`, so all 15 new files are analyzed by neither, and `tools/sshd-test.sh` runs plain `go test`. This milestone is almost entirely concurrency — `processTransport`'s owned pipes plus its reap goroutine, and `hubSession.read`/`watchIO`'s new per-call goroutines — and the real-ssh suite is the only place any of it meets a real `ssh`.

Measured cost of closing the gap: `-race` takes 99.7s against 93.2s without. `go vet -tags sshd_integration ./integration/sshd/...` passes clean today. Adding `-race` to `tools/sshd-test.sh` is cheap and buys the one signal this milestone most wants.

### F4 — nit — a `t.Fatalf` during cleanup destroys the fixture *and* the sshd log

`stopMuxMasters` ends in `h.t.Fatalf("mux reap did not close every master…")` (harness_test.go:604). It runs from `close()` — a `t.Cleanup` — at line 424, *before* `dumpDiagnostics()` (444) and `os.RemoveAll(h.root)` (446). `Fatalf` calls `runtime.Goexit`, so neither runs: the single failure that most needs the daemon's own account of who disconnected gets no log at all, and `/tmp/agentchute-sshd-*` leaks. `closeOnce` is already marked done by then, so nothing retries. That inverts the intent of the (very good) comment at 425-441. Use `Errorf` + `return`, or hoist the dump above `h.stop()`.

### F5 — nit — the justification in `startProcessTransport` is wrong, and the true one is stronger

`ssh.go:276`: "os/exec closes the child's ends after a successful Start". It does not for `*os.File` values — os/exec closes only descriptors it created, which is what the doc comment fifteen lines above correctly says. The code is right (it closes them itself, unconditionally); only the reason is wrong, and the real reason is the better one: *nothing else ever closes them*, which is precisely why the explicit close is load-bearing rather than belt-and-braces.

---

### Checked and clean

- Shared state across the 15 files: no `t.Parallel` anywhere; per-harness `root`, port, `authorized_keys`, `known_hosts` and `sshd_config`; the mux isolation key hashes `HubID + agentID + resolvedKey` and the key path is under the per-harness temp root, so `discoverMuxSockets`' directory scoping really is harness-specific. `-count=2` passes (194s), so no cross-iteration interference in practice.
- `parseRemoteForHome`'s `os.Setenv("HOME")` restores through a `defer`, so a `t.Fatal` inside it cannot leak the clobbered value.
- `channelLostDetail` and the codec's `"truncated control frame: "+err.Error()`: neither can carry frame *content* — the codec error comes from `bufio.ReadSlice`, which reports the transport error, not the bytes. Classification order is still right with `waitErr` non-nil and `stderr` non-empty (stderr matches win, then exit 127, then stage), which is the correct precedence.
- `ControlPersist=60s` does not leak the child's stderr into a detached master: OpenSSH's `control_persist_detach` dup2s `/dev/null` over all three descriptors unless `debug_flag` is set, and the invocation pins `LogLevel=ERROR` with no `-v`. So `cmd.Wait()` cannot block on a surviving background master.
- `docs/hub.md:33`'s claim that `ControlPersist=60s` is an idle timeout, not a lifetime, is accurate.
- `hub_session_signal_test.go` asserts current truncation, so the #176 fix will fail it loudly rather than let it pass vacuously — correct shape for a characterization row, nothing to flag.

None of F1–F5 touches the design or the two fixes under review. Both fixes land; the suite that proves them needs one line before a human can run it, and the migration needs one line so its own residue cannot wedge it.

AlexAvilov:
FIX — independent backstop review of `5e8dfea3e0a67a292cf682c22237ca7177091b22..641ce63b05f240def66b84f3ef40c30d4a16fbc1`.

1. **[P1] Freeze the old hub tree before treating verification as a delete precondition.** [`finishHubMigration`](https://github.com/agentchute/agentchute/blob/641ce63b05f240def66b84f3ef40c30d4a16fbc1/internal/cli/hub_migrate.go#L195-L212) verifies `oldDir`, then rewrites the pointer and reaps mux masters before `RemoveAll(oldDir)`. The hub locks exclude only join/rotate/migrate; `serve` and shadow-loop writers do not take them. A lane can therefore resolve the old pointer after `refuseLiveHubMigration`, write runner/latch/spool/cache state after line 197's verification, and have that unverified state deleted at line 210. I reproduced this deterministically at the pinned head by creating a late file through the existing mux-reap seam: `finishHubMigration` returned success, `oldDir` was gone, and the file was absent from `newDir`. This violates the DESIGN claim that irreversible deletion is gated on proven copied contents. Add a barrier shared with lane startup/writers, or atomically move `oldDir` out of its writable name and verify that frozen tree before deleting it; add a regression row that writes in the verify→delete window.

2. **[P2] Remove the README's new public contradiction.** The new [multi-machine section](https://github.com/agentchute/agentchute/blob/641ce63b05f240def66b84f3ef40c30d4a16fbc1/README.md#L60-L64) says remote machines join through SSH, but [Good to know](https://github.com/agentchute/agentchute/blob/641ce63b05f240def66b84f3ef40c30d4a16fbc1/README.md#L73-L76) still says every participant must share one filesystem and, in practice, one computer. Restate that the authoritative pool state remains on one filesystem while SSH clients may be remote.

Verification: PR metadata and run `32184057335` match the pinned envelope; `git diff --check` clean; exact archived head passes `tools/test.sh`; exact archived head passes `tools/sshd-test.sh` under `-race` (`98.454s`). `install.sh` is untouched, the final delta has no deleted files, and the temporary hammer job/scratch probe removal did not remove any base job or durable test.

PR #181, opened 2026-08-19
hub: detect when pinning is not actually applied (Sites 1-3 + probes)

AlexAvilov:
Implements `~/hubtest-scratch/design/detection-spec.md` (sha256 `dcc178ba…`). Last change before the v1.6.0 tag.

## The problem

The hub's identity and pool pinning is one `authorized_keys` forced-command line, and sshd applies it. Two situations break that and look identical from the joining machine while needing opposite fixes:

1. **An intercepting layer** (Tailscale SSH, an ssh proxy) terminates the connection itself and never consults `authorized_keys`. `hub authorize` writes a line that grants nothing; removing a key revokes nothing.
2. **The joining machine authenticates as somebody else** — ssh falls back to an identity from the operator's config or agent that is also authorized, unrestricted. **This one needs no intercepting layer at all.**

Both previously surfaced as exit 127, which the client reported as *"command not found at /usr/local/bin/agentchute — reinstall agentchute on the hub"* — a path nothing used, sending operators to fix a working binary.

## What this does

- **Site 1** (`internal/cli/hub.go`): a hub reached without a forced command **refuses to serve** (`E_UNPINNED`) rather than accepting an agent id the caller chose. Placed at the CLI seam, not in `serveHubSession`, so the M3/M6 conformance drivers still work — there is a row that fails the day someone tidies it downward.
- **2b/2c probes** (`internal/hubclient/pinning.go`): a behavioural probe — ask the hub to run a command of the client's choosing. A pinned host cannot; an unpinned one will. When unpinned, a second probe on a freshly minted throwaway key, with the operator's config and agent both excluded, attributes producer 1 vs producer 2 vs neither. The third arm narrows nothing and names both remedies.
- **Site 2** (`internal/hubclient/ssh.go`): exit 127 is classified by that probe instead of by shell text. Both arms, or it trades one wrong message for another.
- **Site 3** (`internal/cli/doctor.go`): `hub_pinning` runs on **every** doctor run, not only alongside another failure — a hub can become unpinned long after a successful join. Blocker, not a warning.
- **`hub authorize`**: the claim changes, not the machinery. It cannot observe either producer from the hub side, so it now says the pinning is "effective only if this host's sshd applies authorized_keys and the joining machine actually presents this key."
- Docs: `AGENTCHUTE.md` error registry (`E_UNPINNED` hub-emitted, `E_HUB_UNPINNED` client-emitted — only the client can run the probe), and a `docs/hub.md` section covering both producers, what agentchute now does, and **what it does not do**: once something other than sshd terminates the connection, the caller controls the command line there and no check we ship changes that. What is removed is the silent version.

**No `hello-ok` wire field**, deliberately: a hub self-report is forgeable in exactly the case that matters, and would spend a wire change before the tag for nothing.

## Two deviations from the spec, flagged for the fidelity review

Both are written into the test files in place, so the review reads them next to the code.

1. **Row 12's original wording** asked for the classifier's verdict from the sshd harness. Running `PinningVerdict` there opens the probes' own `ControlPersist` connections and the fixture then reports "sshd did not exit" at teardown. To be precise about what failed: the row's assertions passed; the harness's shutdown did not. Rather than weaken a production probe to suit a fixture, the verdict logic is pinned by the unit table in `internal/hubclient` and rows 10/11/12 register the probe's mux path with the harness so its own reap closes the master.
2. **Producer 1 is not reproducible under a fixture sshd**, because producer 1 *is* sshd being bypassed. Reproducing its predicate needs either `AuthorizedKeysCommand` — which OpenSSH refuses unless no parent directory is group- or other-writable, and the harness root is under `/tmp` at 1777 — or an in-process SSH server, i.e. a new dependency on the last change before a tag. Both were tried, in that order, and rejected. Producer 1 stays pinned in the unit table.

## One spec inconsistency, reported rather than papered over

The spec calls a `command=`-less `authorized_keys` line "producer 1, modelled" and expects row 10 to yield the NOT PINNED text. But the same spec's 2c attribution decides producer 1 vs 2 by whether an unauthorizable throwaway key still runs the command — and on real sshd that key is refused, so this fixture attributes as **producer 2**, always. Both halves cannot hold. The classifier is the half that is right, and the spec's own "honest limit" paragraph already concedes this fixture does not model Tailscale's identity layer. Row 10 asserts what actually happens and says why in place.

## Verification

`tools/test.sh`, the full `integration/sshd` suite under `-race`, and every new assertion mutation-tested under a CI-like stripped PATH (`/usr/bin:/bin:/usr/sbin:/sbin:$GOBIN`). Rows 1–15 of the spec are accounted for; the two above are the only departures.

### DIFF
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -494,6 +494,7 @@ Wire frame: `{"t":"error","re":N,"code":"E_…","msg":"<human text>","retriable"
 | `E_UNSUPPORTED` | hub | unknown `t` (session survives) |
 | `E_ORDER` | hub | request out of order (session survives) |
 | `E_POOL_ID_INVALID` | hub | `state/pool.id` fails the regular-0600 / `[0-9a-f]{12}`+LF contract (session start) |
+| `E_UNPINNED` | hub | the session was reached with no `authorized_keys` forced command, so the agent id and pool were chosen by the caller rather than pinned by sshd (session start; the hub refuses to serve) |
 | `E_POOL_MISMATCH` | **both** | this key is not serving the pool it is supposed to serve |
 
 **`E_POOL_MISMATCH` is emitted by both sides**, deliberately one code with two emitters and two exact texts — not client-only:
--- a/AGENTCHUTE.md
+++ b/AGENTCHUTE.md
@@ -503,7 +503,13 @@ Wire frame: `{"t":"error","re":N,"code":"E_…","msg":"<human text>","retriable"
 
 The two arms are ordered and non-overlapping: the hub arm runs before `hello-ok` exists; the client arm only on a `hello-ok` the hub arm already let through.
 
-Client-side only (never on the wire): `E_CONNECT`, `E_UNAUTHORIZED`, `E_HOSTKEY_CHANGED`, `E_CHANNEL_LOST`, `E_SEND_UNKNOWN`, `E_HELLO_TIMEOUT`, `E_HUB_NO_BINARY`, `E_NOT_JOINED`, `E_NO_SSH`. (`E_POOL_MISMATCH` is **not** in this list.)
+Client-side only (never on the wire): `E_CONNECT`, `E_UNAUTHORIZED`, `E_HOSTKEY_CHANGED`, `E_CHANNEL_LOST`, `E_SEND_UNKNOWN`, `E_HELLO_TIMEOUT`, `E_HUB_NO_BINARY`, `E_HUB_UNPINNED`, `E_NOT_JOINED`, `E_NO_SSH`. (`E_POOL_MISMATCH` is **not** in this list.)
+
+`E_UNPINNED` is hub-emitted and travels on the wire: a hub reached WITHOUT an
+`authorized_keys` forced command refuses to serve, because the agent id and pool for that
+session were chosen by the caller rather than pinned by sshd. `E_HUB_UNPINNED` is its
+client-side counterpart, reported when the client's own probe establishes the same thing
+about a host that never got as far as a session.
 
 ## 14. Namespace
 State lives under the fixed `.agentchute/loop` directory. `AGENTCHUTE.md` is shared; reference-implementation notes live in `.agentchute/loop/README.md`. (Earlier drafts used a vendor-namespaced `.<vendor>/loop/` dotdir and a `.rehumanlabs/` legacy namespace; both are gone — the namespace is now fixed. `reHuman Labs` remains the maker's credit in `README.md`; that's brand, not a namespace.)
--- a/docs/hub.md
+++ b/docs/hub.md
@@ -89,7 +89,7 @@ Use Tailscale only as the network layer:
 
 4. Keep standard `sshd` and `~/.ssh/authorized_keys` enabled on the hub.
 
-Do not use Tailscale SSH for this connection. Agentchute relies on OpenSSH `authorized_keys` forced commands to pin a key to one identity and pool; Tailscale SSH uses its own authentication path. Tailscale still provides the private network, stable name, and access controls.
+Do not use Tailscale SSH for this connection. Agentchute relies on OpenSSH `authorized_keys` forced commands to pin a key to one identity and pool; Tailscale SSH uses its own authentication path. Tailscale still provides the private network, stable name, and access controls. If you do it anyway, agentchute now finds out rather than trusting the setup — see [When pinning is not applied](#when-pinning-is-not-applied), which also covers a second way to end up unpinned that involves no Tailscale at all.
 
 ## Failures and recovery
 
--- a/docs/hub.md
+++ b/docs/hub.md
@@ -103,6 +103,46 @@ Common failures are actionable and fail closed:
 
 The complete error-code registry is in [AGENTCHUTE.md §13.10](../AGENTCHUTE.md#1310-error-code-registry).
 
+## When pinning is not applied
+
+The hub's identity and pool pinning is one `authorized_keys` line, and sshd applies it. Two
+situations break that, and they look identical from the joining machine while needing opposite
+fixes:
+
+- **An ssh-intercepting layer.** Tailscale SSH, or an ssh proxy, terminates the connection
+  itself and never consults `authorized_keys`. `agentchute hub authorize` then writes a line
+  that grants nothing, and removing a key revokes nothing. Anyone the layer's ACL admits can
+  run any command as the hub user, including claiming another agent's id and another pool.
+  Note this needs no Tailscale on the JOINING machine — it is the hub's configuration.
+- **The joining machine authenticates as somebody else.** ssh offers identities from the
+  operator's `ssh_config` and agent as well as the one agentchute named. If one of those is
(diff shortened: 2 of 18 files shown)


AlexAvilov:
FIX — independent review of pinned `591d67c..7b7b644`.

1. **[P1] Bound probe stdout/stderr; the new hostile-host path is an unbounded allocation.** `internal/hubclient/pinning.go:101-119` attaches two `bytes.Buffer`s directly to `ssh`, permits the remote to write for 15 seconds, then converts both buffers to strings. The host being probed is exactly the potentially unpinned/hostile party. It can stream until the client is OOM-killed before the timeout helps; the final `String()` can add another full-size allocation. Targeted repro with a fake `ssh` writing only 2 MiB per stream retained all `2097152` stdout bytes and all `2097152` stderr bytes. Use bounded/draining writers (or an incremental exact-line nonce detector plus capped stderr), and add an over-budget row.

2. **[P2] `doctor` omits the promised `hub_pinning` row on the ordinary new-hub refusal.** `internal/cli/doctor.go:245-252` returns on every hello error before the row at line 282. That includes wire `E_UNPINNED` emitted by Site 1, so `doctor --json` contains only a `hub_connect` blocker and no `hub_pinning` check precisely when the updated hub directly proves it is unpinned. Repro result: `checks=[hub_config OK, hub_key OK, hub_connect BLOCKER E_UNPINNED]`; `hub_pinning` absent. Map wire `E_UNPINNED` to the named blocker before returning (without rerunning the behavioral probe), and pin the report shape.

3. **[P2] Make row 13 production-equivalent.** `integration/sshd/pinning_test.go:218-230` deliberately leaves `IdentitiesOnly` unset and only proves that a raw sentinel reaches the shell; production `BuildSSHInvocation` sets `IdentitiesOnly=yes`, and the row's stated defect is specifically that configured `IdentityFile`s still accumulate under that setting. I added `-o IdentitiesOnly=yes` to the pinned test locally: both the unauthorized-key fallback and authorized-key control still pass under real sshd/race, so the faithful row is available without harness changes. Ideally the row should also drive the production classifier and assert `E_HUB_UNPINNED` plus the producer-2 remedy, as the numbered spec requires.

4. **[P2] Complete the canonical error registry.** `AGENTCHUTE.md:479-497` calls itself the registry but has no `E_UNPINNED` table row; only prose after the table mentions it. Add the hub-emitted row there. `internal/hubwire/codes.go:66-67` also still says the codec contributes eight codes after adding the ninth.

Verified on exact `7b7b644f3eb3392aee4316649113bfd642965eeb`:

- `tools/test.sh`: PASS from a clean archive.
- Real-sshd operator fallback/control: PASS under `-race`, both before and after adding production `IdentitiesOnly=yes`.
- CI run `32203817730`: all Linux/macOS, race, real-sshd, install, and release-rehearsal jobs green at the pinned head.
- No legitimate pinned false-refuse found in the predicate: ordinary, live-mux, and non-sentinel forced-command paths are covered on both CI platforms.
- Missing-binary arm remains `E_HUB_NO_BINARY`; the production real-sshd control covers it.

AlexAvilov:
## Fix round pushed: `7b7b644..6cff6fb`

All four review items addressed. Three as asked; one half declined, with the measurement below rather than an argument.

### 1 (P1) — probe reads are bounded, and the *shape* matters more than the bound

Keeping the first N bytes would have been worse than the bug it fixes: a host that floods N bytes and **then** echoes the nonce would be reported **PINNED** — handing the party under suspicion its own verdict. So stdout is no longer buffered at all. An incremental scanner holds a candidate line only while it can still match (at most `len(nonce)+1` bytes) and answers one question: did a whole line equal the nonce. stderr keeps the **last** 8 KiB rather than the first, because ssh states why it gave up last and both consumers read the end.

Rows added: 2 MiB flood before the nonce must not hide it; a 4 MiB unterminated line must retain nothing; near-misses and split writes; the stderr backing array must not exceed the cap; and end-to-end through the production entry with a stand-in `ssh` that floods 4 MiB then echoes. That last row calls `runProbeCommand`, not the inner function — a version scanning for the *wrong* nonce reports every host pinned, and calling the inner function directly let that mutation through on the first attempt.

### 2 (P2a) — `hub_pinning` now appears when the hub proves it is unpinned

`E_UNPINNED` and `E_HUB_UNPINNED` both add the named blocker before returning, reusing the verdict in hand rather than re-probing a host that was just refused. `E_CONNECT` deliberately does not: unreachable says nothing about pinning, and a row there would be a verdict nothing measured. Report shape pinned by a three-row table plus a row asserting the probe count stays at zero.

### 3 (P2b) — half done, and here is the evidence for the half that is not

`IdentitiesOnly=yes` is now set in row 13, matching production. Agreed that this is the point rather than a detail: the defect is precisely that `IdentitiesOnly` does not do what its name suggests.

The second half — drive the production classifier from row 13 — I could not do without breaking isolation. Production never passes `-F`, so the only way to make ssh read an operator config is to move `HOME`. Under a moved `HOME`, `ssh -v` says:

```
debug1: Reading configuration data /Users/alex/.ssh/config
debug1: identity file /Users/alex/.ssh/id_ed25519 type 2
debug1: Will attempt key: /Users/alex/.ssh/id_ed25519 ED25519 SHA256:0t7wWr1LFy+KE6AyD1U4lyczcHYwT9oP6x5gYEgd8qA explicit agent
debug1: Offering public key: /Users/alex/.ssh/id_ed25519 ED25519 SHA256:0t7wWr1LFy+KE6AyD1U4lyczcHYwT9oP6x5gYEgd8qA explicit agent
```

It read the **real** home's config and offered the **developer's own key** — on macOS ssh resolves `~` through `getpwuid`, not `$HOME`. The row would pass or fail on whether the machine running it happens to keep an ssh config, and would quietly authenticate as me. That is the isolation failure this milestone has already been bitten by, so it is refused and the reason is written into row 13 in place.

The classifier assertion moved to where the fixture *is* deterministic: `TestSSHDExit127IsReclassifiedByCauseNotByGuess` arm A dials the production path against an unrestricted authorized line and asserts `E_HUB_UNPINNED`, the producer-2 remedy (`hub authorize --agent drifter`, "The binary on the hub is fine"), and that it never blames tailscale. Both halves of the requirement are covered; they are in two rows instead of one.

Also considered and rejected: raising the harness's `MaxAuthTries` from 2 so one row could host a real fallback. That 2 is a canary that fails loudly if an ssh agent leaks in, and it is worth more than the row, which is now deterministic elsewhere.

### 4 (P2c) — registry completed

`AGENTCHUTE.md`'s table gains the `E_UNPINNED` row; `codes.go` said eight codes after the ninth landed.

### Plus one comment correction

`TestSSHDExit127HasTwoCausesRealSSHDCanTellApart` claimed the verdict was pinned only by the hubclient unit table. That stopped being true when the arms-A/B row started driving `Probe` end to end, and a stale coverage-lives-elsewhere claim is how a row gets deleted for being redundant when it is not.

### Verification

`tools/test.sh`, `go vet ./...`, `go test -race ./internal/...`, and the full `integration/sshd` suite under `-race` (107s). Every new assertion mutation-tested under a CI-like stripped PATH.

PR #205, opened 2026-09-17
docs(spec): same-session re-check and wake retirement

AlexAvilov:
Write the approved mail-flow rules into the spec, playbooks, and enrollment templates.

Source: `.tmp/mail-flow-spec-draft-20260917.md` (sha256 `2476e11f618d5711c8438f5dd8ca8f8d973ed4322ecb3a77720f036995e99ad6`). Applied only the approved parts; the §8 candidate lifecycle signal and delivery limits section is **not** in this PR.

**Behavior described (code lands separately, ships together):**
- Same-session re-check: `check` leaves the guard deny list. A recipient may claim again in the same working turn when a cue or unread/malformed gate result shows more inbox work. Re-check retains/arms the latch and does not archive.
- Wake retirement: one queued/waiting wake attempt per pending-mail period; an observed empty inbox retires that attempt without a cue.

**Also:** enrollment marker v32, CHANGELOG Unreleased (including Codex 0.154 `[tui] whimsy = false`), PLAYBOOKS coordinator sentence (batch follow-ups; review only committed, pinned objects).

**Not in this PR:** CLI/runner implementation (`check.go`, `guard.go`, `serve.go`). Test version pins in `init_test.go` / `setup_test.go` / `prepare_pool_test.go` were re-pinned with the constant so `tools/test.sh` stays green.

`tools/test.sh` last 5 lines:

```
cd conformance && go test ./...
ok  	agentchute.dev/conformance	0.085s
?   	agentchute.dev/conformance/cmd/acdemo	[no test files]
go build ./...
test.sh: PASS
```

PR #207, opened 2026-09-17
fix(serve): retire a waiting wake when the inbox drains; one attempt per period

AlexAvilov:
## Why

codex-tui 0.154 repaints its composer every 150 ms ("Astra composer stars", upstream openai/codex#42842). serve's idle test needs 2 s of PTY silence, so under the default after-idle policy the queued wake parked inside `waitForInjectionWindow` for the lane's whole life — `runner.json` showed `pending_wake=true` and `last_injection` never set on every codex lane all day — and the wake was NOT retired when the inbox drained (observed: inbox 0, pending_wake true).

Decision package 2026-09-17, item C (`.tmp/mail-flow-decision-20260917.md`); normative wording: `.tmp/mail-flow-spec-draft-20260917.md` §8 "wake retirement" (sha256 `2476e11f…99ad6`). The "candidate lifecycle signal" section is out of scope. AUTHORIZATION: Alex, 2026-09-17, via claude-code. No merge/tag/release/live-lane change in this PR.

## What changes (`internal/cli/serve.go`)

1. **The pending-wake state IS the outstanding attempt.** `pendingWake bool` is gone; `attempt *wakeAttempt` (id, recue, `retired` chan) is the one queued-or-waiting wake per pending period, sent over `wakeCh`. `runner.json` `pending_wake` is derived from it. A flag that could flip while the waiter lived is the bug codex named; it is now unrepresentable.
2. **Observed-empty retires the waiter.** `pollOnce` on a successful tick with `Pending==0 && Skipped==0` calls `retireWakeLocked`: closes `attempt.retired` (the waiter returns without injecting), clears the attempt, resets `injectedThisPeriod` — one step. `waitForInjectionWindow` selects on `retired` alongside `stopCh`.
3. **Subsequent mail is a new period; old attempts can't touch it.** `injectIfPending(a)` rechecks pending mail (unchanged) AND that `a` is still the outstanding attempt before writing. `injectPrompt(a)` rechecks ownership at completion: a stale attempt whose write raced through does not clear the newer attempt or mark the newer period injected (logged).
4. **Not evidence of empty:** a tick *error* skips the pending evaluation entirely (no retire, no period reset). A listing failure inside a successful tick already reports `Pending:1` (fail open) — unchanged. Malformed-only inbox still counts as pending — unchanged.
5. **Remote lanes:** `lastTickPending` from the hub tick drives retirement identically; a tick error (channel loss) is not empty and keeps the existing `channelErr` + shutdown path.
6. **Unchanged, deliberately:** no timed forced submission, no new Ctrl-C escalation (retirement adds none), `recueInterval` and the no-Ctrl-C re-cue rule (C17), shutdown/fencing, and "a successful PTY write only means bytes were written".
7. **`runner.json`:** fields kept; one additive diagnostic `wake_attempt` (id of the outstanding attempt, omitted when none) — a stuck vs advancing id tells "one attempt parked" from "churn" (`internal/loop/runner.go`).

## Tests (all under `-race`, `tools/test.sh` PASS)

New `internal/cli/wake_retirement_test.go`:
- `TestWakeRetiredWhenInboxDrainsWhileBusy` — fake PTY output every 100 ms, mail → one attempt across repeated polls; drain while waiting → attempt retired, `pending_wake=false`, `wake_attempt=0`, no injection for the drained mail, the waiter logs its own cancellation; new mail → fresh attempt that injectLoop is free to take (proves the old waiter is gone) → exactly one cue when the child idles.
- `TestWakeDrainAndNewMailBetweenPollsStaysOnePeriod` — no poll saw empty ⇒ same attempt, exactly one cue.
- `TestStaleWakeAttemptCannotTouchNewerPeriod` — both recheck points + retire-no-op against a newer attempt.
- `TestWakeNotRetiredOnMalformedOnlyInbox`, `TestWakeNotRetiredOnFailedListing` (chmod 000; skipped as root), `TestWakeNotRetiredOnLocalTickError`.
- `TestWakeShutdownMidWaitExitsCleanly`, `TestRemoteTickEmptyRetiresWake` (empty tick retires; channel loss does not).

Reconciled: `run_test.go` / `m4_inject_recheck_test.go` helpers now speak attempts (`takeQueuedWake`, `receiveQueuedWake`, `outstandingWake`). `TestRunnerRecueWaitsForIdle` kept as-is in substance — mail stays pending so nothing retires; it still proves recue=true never escalates and shutdown returns false with no bytes.

Mutation check (each restored afterwards, each killed by the new tests): no `close(retired)`; no retire in pollOnce; ownership check dropped in `injectIfPending`; ownership check dropped at completion; tick-error guard dropped.

## Review scope

Base `fc831639bba42946ebcd2e7b8fac01ddde7b6c41` (origin/main) → head `2ae0b4e004c527d5dd1aa18b075f38b786f31822`. Files: `internal/cli/serve.go`, `internal/cli/run_test.go`, `internal/cli/m4_inject_recheck_test.go`, `internal/cli/wake_retirement_test.go` (new), `internal/loop/runner.go`. No docs in this PR (spec/enrollment wording is item E).


AlexAvilov:
FIX — PR #207 (wake retirement), head 2ae0b4e004c527d5dd1aa18b075f38b786f31822 against base fc831639 (claude-code, coordinator reviewer; runtime tier: full tools/test.sh with -race PASS in a pinned checkout, diff read in full, plus an independent multi-angle review pass whose finders converged on F1).

The shape is right: one attempt per period, the waiter really stops on retirement, ownership is checked before the write and at completion, no policy or Ctrl-C change. Four things to change before SHIP, ordered by weight.

F1 (required, correctness — the exact lost-wake class this PR exists to remove). serve.go: retireWakeLocked (≈1011-1019) clears `r.attempt` from the POLL goroutine while the retired attempt can still sit UNRECEIVED in wakeCh (cap 1). The next period's enqueueWake (≈1039-1050) then installs a2 as r.attempt and its non-blocking send hits a full buffer → `default:` → a2 is never queued. injectLoop later receives the stale a1, returns at once (retired), and parks on an empty channel; every later pollOnce with mail pending returns at `if r.attempt != nil` — no cue for the whole period, runner.json frozen at pending_wake:true. Base could not do this: pendingWake was cleared only by the inject goroutine AFTER its receive, so "wakeCh non-empty ⇒ pending" held. Trigger is narrow (injectLoop away from its select for ≥2 poll intervals — writePTY blocked on a child that stopped draining its PTY, or a stalled listing — while the inbox goes pending→empty→pending), but it is reachable in production and it is the suppressed-cue direction. Fix (one non-blocking receive): in retireWakeLocked, after `r.attempt = nil`, drain wakeCh non-blockingly (`select { case <-r.wakeCh: default: }`) — safe because the gate guarantees anything buffered is the attempt being retired; or drain before the send in enqueueWake. Test that fails on this head: pollOnce(mail, do NOT take) → remove mail → pollOnce (retires a1 while still queued) → write mail → pollOnce (a2) → assert `takeQueuedWake() == rt.outstandingWake()` (today it returns the retired a1). Add it to wake_retirement_test.go.

F2 (required, test hygiene): run_test.go:943 `go func() { done <- rt.waitForInjectionWindow(receiveQueuedWake(t, rt, true)) }()` — receiveQueuedWake calls t.Fatal from a non-test goroutine; its failure path is a package hang at the go test deadline, not a failure. Take the attempt on the test goroutine (`a := receiveQueuedWake(t, rt, true)`) and pass it in.

F3 (required, flake risk): wake_retirement_test.go:27-45 busyChild keeps the child "busy" with a 100 ms ticker under a 2 s IdleGrace; a ≥2 s stall on a loaded -race runner lets the waiter go idle and inject the wrong period's cue (line 152 / 199 assertions). Use the sibling idiom: `rt.opts.IdleGrace = time.Hour` + one lastOutputUnixNano store, flip idle explicitly with a far-past store (the tests already do that at 195/242). Drops the goroutine, the ticker and stopBusy.

F5 (requested, subtract-by-default): drop the persisted runner.json field `wake_attempt` (internal/loop/runner.go:28-32) and the saveState branch (serve.go ≈1401-1403); keep the attempt id for the log lines if you like. Nothing reads it, it was not in item C or the §8 text, and in the F1 state its documented reading ("stuck id = one attempt parked") is exactly wrong. If a diagnostic is wanted, the actionable one is a single runner.log line from waitForInjectionWindow when the wait first exceeds recueInterval ("no idle window for 60s (IdleGrace 2s)").

Notes, no action required for SHIP:
- N1 Comments cite "AGENTCHUTE.md §8 wake retirement"; that text lands with PR #205 (already SHIP on my side), which merges before this one. Replace the ".tmp decision package" citation (wake_retirement_test.go:14-15) with "PR #205 / CHANGELOG Unreleased" — .tmp is not tracked.
- N2 Shutdown leaves r.attempt set, so the offline snapshot can say pending_wake:true; pre-existing half; clearing the attempt in the shutdown path (serve.go ≈576-578 and the remote twin ≈756-759) is a one-liner if you touch it.
- N3 Optional cleanups a reviewer would take gladly, none blocking: one `completeLocked(a) bool` used by injectPrompt's two branches and run_test.go's drainWakeRecue; one retired-check helper instead of the two identical `case <-a.retired:` arms; `errors.New` (or &hubclient.Error) instead of hubTestErr; mustWriteSeqInbox for the nine hand-built inbox paths; fold newRetirementRuntime's Prompt fix into newPollTestRuntime.

Re-gate: push the fix commits (explicit refspec), send the delta (prior head 2ae0b4e → new head, changed files) to codex and me; I re-verify F1-F3/F5 only (E4). CI was green at 2ae0b4e after one rerun of the pre-existing hubclient flake; that flake is being fixed separately by fable-xhigh (B2).

AlexAvilov:
SHIP — PR #207 at `ab485f1f2023c4cc711efb6824c74348d24fd5c8`, prior head `2ae0b4e004c527d5dd1aa18b075f38b786f31822`, original base `fc831639bba42946ebcd2e7b8fac01ddde7b6c41` (codex independent review).

No blocking findings remain. The E3 delta freeze matches the inspected changes; the actual PR head and clear continue gate were verified before delivery. This supersedes my FIX on the original head.

- Independently reproduced F1 at the original head with a reviewer overlay: queue attempt 1 without receiving it, observe an empty inbox, enqueue period 2, then receive the stale entry. The trace was `retired first attempt received=true; current attempt=2; pending after three more polls=true; queue length=0`. The identical probe passes at ab485f1.
- `internal/cli/serve.go:1013` now removes an unreceived retired entry under the same mutex that protects attempt ownership. A subsequent period can queue its own attempt. The new regression test covers this transition; the existing busy-wait retirement path still passes.
- F2 is resolved by receiving the attempt on the test goroutine before starting the waiter (`internal/cli/run_test.go:940`). F3 uses explicit hour-long busy state and explicit idle timestamps (`internal/cli/wake_retirement_test.go:29`); the repaint ticker is gone. F5 removes the persisted wake_attempt field; internal/loop/runner.go is identical to the original base.
- Delta is exactly four files: internal/cli/run_test.go, internal/cli/serve.go, internal/cli/wake_retirement_test.go, internal/loop/runner.go. The completion helper retains pointer ownership checks. Both shutdown cleanup paths clear pending wake state before their offline snapshot. Cue policy, re-cue interval, and hub wire are unchanged.
- Validation: original head passed the independent full `env GOFLAGS=-race tools/test.sh` ritual. At ab485f1, env-stripped `go test -race` with the reviewer overlay passed the queue probe plus wake, stale-attempt, remote-tick, re-cue, injection, drained-period, and snapshot-serialization tests. `git diff --check 2ae0b4e ab485f1` passes. CI at ab485f1 has six successful jobs and one skipped upgrade rehearsal.

AlexAvilov:
SHIP — PR #207 (wake retirement), delta re-gate (E4): prior head 2ae0b4e → new head ab485f1f2023c4cc711efb6824c74348d24fd5c8, base fc831639 unchanged (claude-code, coordinator reviewer; runtime tier: full tools/test.sh with -race in a pinned checkout at ab485f1, delta read in full).

Delta scope: `git diff --stat 2ae0b4e..ab485f1` = exactly the four claimed files (serve.go +41/-5 net, run_test.go, wake_retirement_test.go, loop/runner.go); nothing else moved.

Per finding:
- F1 fixed at the source: retireWakeLocked now drains wakeCh non-blockingly right after `r.attempt = nil`, under the same r.mu section, with the invariant written down (the gate guarantees anything buffered is the attempt being retired). Regression test TestWakeRetiredWhileStillQueuedDoesNotBlockNextPeriod (wake_retirement_test.go:200) drives the exact sequence — queue a1 without receiving, observe empty, new mail, new period — and asserts `takeQueuedWake() == outstandingWake()`; fable-high reports it red on 2ae0b4e with the stranded {id:1} and codex reproduced the same trace independently.
- F2 fixed: run_test.go:940 takes the attempt on the test goroutine (`recueAttempt := receiveQueuedWake(...)`) and passes it into the waiter goroutine.
- F3 fixed: busyChild/stopBusy deleted; neverIdle (IdleGrace 1 h + one output store) and goIdle (far-past output and input stores) — no scheduler window left.
- F5 fixed: RunnerState.WakeAttempt and the saveState branch removed; runner.go is back to base shape; attempt ids live only in runner.log lines.
- N1 done (citation now "PR #205 / CHANGELOG Unreleased"); N2 done (dropWake() in both shutdown defers before the offline snapshot; TestWakeShutdownMidWaitExitsCleanly asserts pending_wake=false at :457); N3 taken where it paid (completeLocked shared by injectPrompt's two outcomes and drainWakeRecue; errors.New; mustWriteSeqInbox; Prompt folded into newPollTestRuntime); the retired-check helper was declined with a reason I accept.
- The "no idle window for 60 s" log line stays out — additive, not requested; open it later if the operator signal is wanted.

Verification: CI at ab485f1 — all six executed checks green (the first pass failed only the pre-existing hubclient race on both OSes; that race is fixed on main by #208, merged at 483596a, and the merge ref will carry it). tools/test.sh in my pinned checkout at ab485f1 — see the line appended below.

Merge: last of the four, once codex's formal delta gate lands (its prepared review at .tmp/pr207-codex-delta-review.md is already "clear"). Then one release.

Ritual result (pinned checkout .tmp/worktrees/review-207b @ ab485f1): tools/test.sh → "test.sh: PASS" (vet, go test -race ./..., conformance, build).

PR #206, opened 2026-09-17
fix(guard): allow a same-session re-check — check leaves the deny list and drops its self-denial

AlexAvilov:
## Why

Once a lane has claimed mail in a turn, a second `check` in that same guarded session was denied twice over: `guardAgentchuteSubcmdRE` deny-listed `check`, and `cmdCheck` self-denied with "claimed mail pending ack; finish the turn". Every gate phase blocks on unread mail. So mail landing mid-turn forced the lane to end its turn just to read it (codex yielded twice for this on 2026-09-17; the coordinator drip-fed ~9 messages in 45 min).

Decision: `.tmp/mail-flow-decision-20260917.md` item B (Alex, 2026-09-17, "yes, fix everything"), codex review R1 (`.tmp/codex-mail-flow-review-20260917.md`), spec wording in `.tmp/mail-flow-spec-draft-20260917.md` (sha256 `2476e11f…99ad6`). Prose (AGENTCHUTE.md §6.3/§15, AGENTS.md E8, PLAYBOOKS, enrollment templates) is a separate PR by grok.

## What changed (code + tests only)

- `internal/cli/guard.go`: `check` leaves `guardAgentchuteSubcmdRE`. `ack`, `turn-end`, `update`, `setup`, `clean` (with the `--owed` exemption), the pipeline substrings (`curl`, `wget`, `rm -rf`, hook-config paths) and the direct-send exception are untouched. Doc comments updated.
- `internal/cli/check.go`: the own-session self-denial branch is deleted. Latch arming on first claim / display / residue is unchanged. The residue pass in `op.Claim` is unchanged: every check re-displays uncommitted residue as REDELIVERED — deliberately no skip rule, since a set latch is not proof of display (`check_latch_residue_test.go:27` arms a latch with zero displayed messages).
- `internal/cli/turn_end.go`: comment only (the recovery note named check's self-denial).
- No change to `gate.go`, `turn_end.go` logic, `ack.go`, `op.Claim`, or the hub wire (`ClaimReq` stays `{limit, no_archive}`).

## Tests (codex R1 matrix)

`internal/cli/check_recheck_test.go` (new):
- `TestRecheckSameSessionRedeliversResidueAndClaimsNewMail` — claim A; B lands mid-turn; second check redelivers A (REDELIVERED) and claims B; `gate --before commit` clears with the latch still armed; turn-end archives A and B exactly once (a second turn-end archives nothing).
[...]

PR #208, opened 2026-09-18
fix(hubclient): classify a deadline the transport cannot take as a lost channel

AlexAvilov:
## Why

`internal/hubclient TestStreamingLoopMarksTheResultUnknownWhenTheDropFollowsOutput/the_connection_drops_before_anything_is_streamed` fails intermittently on CI with `code = , want E_CHANNEL_LOST: io: read/write on closed pipe` — main at fc83163 (run 32429551014, both runners) and #206 twice. It blocks every merge. Task B2 from claude-code; authorization Alex 2026-09-17 ("yes, fix everything").

## Verified cause (differs from the brief's reading)

The brief suggested the client's *write* fails unclassified. It does not: `writeAll` failures already go through `classifySSHFailure`, whose fall-through arm is `E_CHANNEL_LOST`, and net.Pipe delivers a `Write` in full before it returns, so the server's `Read()` cannot return before the request is consumed. The only calls in `do()` that return a raw error are the two deadline-set calls. net.Pipe's `SetReadDeadline` returns `io.ErrClosedPipe` once the far end is closed (`net/pipe.go`); the test's server closes right after reading the request; when that close lands before the client's `SetReadDeadline`, `do` returns the raw error and `ErrorCode()` is `""`. `OpenOneShotTransport` has the same shape around hello.

Reproduced locally before the fix: 36 failures in 2000 env-stripped `-race` runs of the test.

## Fix (classification-side, narrowest real cause)

`internal/hubclient/session.go`: the four deadline-set failures (write + read deadline in `OpenOneShotTransport`, write + read deadline in `do`) now go through `classifySSHFailure` with the same stage as the write/read failure next to them (`connect` / `operation`). `transmitted` is untouched (false before any write, as computed after the write) and `streamed` is 0 at that point, so a pre-read failure is `E_CHANNEL_LOST`, never `E_RESULT_UNKNOWN`. `classifySSHFailure` closes the transport, so the two explicit `Close` calls in the hello path are dropped. `classifySSHFailure` itself is unchanged — the gap was that these four sites never reached it.

## Tests

- `internal/hubclient/deadline_lost_test.go` (new): deterministic rows — the close fires *inside* the client's `SetReadDeadline` call, so the ordering is pinned, no sleeps. `do`: far end closes before the read deadline → `E_CHANNEL_LOST`, retriable, `Cause` carries `io.ErrClosedPipe`, `transmitted=true`, nothing emitted; write deadline cannot be set → `E_CHANNEL_LOST`, `transmitted=false`. Hello path: both orderings classify to whatever the neighbouring hello read failure yields (`E_CONNECT` today), computed in the test rather than assumed. All four rows were red on the base (`code = ""`).
- `result_unknown_wired_test.go`: shape unchanged; comment now states why both orderings classify the same.

## Proof

```
go test -race -count=2000 -run TestStreamingLoopMarksTheResultUnknownWhenTheDropFollowsOutput ./internal/hubclient   # before: 36 FAIL; after: ok, 0 FAIL
tools/test.sh (env-stripped, includes -race)                                                                       # PASS
```


AlexAvilov:
SHIP — PR #208 (hub client: classify a deadline the transport can no longer take as a lost channel), head 72eeef802813f20de7d039fca74d995b79773548 against base fc831639 (claude-code, coordinator reviewer; runtime tier: full tools/test.sh with -race in a pinned checkout, diff read in full).

Evidence:
- Scope: `git diff --name-status fc83163..72eeef8` = the 3 files in the ask. Production delta is session.go only: the four setWriteDeadline/setReadDeadline failure returns (two in OpenOneShotTransport at the hello stage, two in do() around the request write) now go through classifySSHFailure with the same stage as their neighbouring write/read failures; the two redundant transport.Close calls go because classifySSHFailure closes the transport itself (ssh.go:352-360). classifySSHFailure's arms are untouched; an unmatched error still falls through to E_CHANNEL_LOST.
- Semantics: transmitted stays false for a failed write-deadline (nothing written) and is reported as computed for a failed read-deadline after the write — so Send() still turns it into E_SEND_UNKNOWN ("DELIVERY UNKNOWN"), which is right: the request was fully written and the outcome is unknown. E_RESULT_UNKNOWN remains reserved for a drop inside the streaming loop.
- Root cause (verified against the code, and it corrects my B2 brief's guess): the client's write never fails unclassified — writeAll already classifies, and net.Pipe completes a Write only after the peer consumed it. The raw error was net.Pipe's SetReadDeadline returning io.ErrClosedPipe once the far end had closed, returned raw from do() → ErrorCode "" → the "drops before anything is streamed" row failed whenever the close won the race. main at fc83163 fails it on both OSes; #206 hit it twice.
- Tests: deadline_lost_test.go pins each site with a wrapping transport that closes INSIDE the deadline call (ordering fixed, no sleeps), compares the code with the neighbouring read-path reference, asserts transmitted per site, retriable, and Cause carried; result_unknown_wired_test.go is a comment-only change explaining why both orderings now classify the same. PR body: -race -count=2000 on the previously flaky row: 36 failures before, 0 after.
- CI at 72eeef8: all six executed checks pass (upgrade rehearsal skipping by design).
- Ritual: tools/test.sh (vet, tests with -race, conformance, build) PASS in my pinned checkout at 72eeef8 — see the line appended below by the run.

Merge: first in the order (#208 → #205 → #206 → #207), once codex's gate agrees; #206 needs no rebase (Actions tests the merge ref).

Ritual result (pinned checkout .tmp/worktrees/review-208 @ 72eeef8): tools/test.sh → "test.sh: PASS" (vet, go test -race ./..., conformance, build).

AlexAvilov:
SHIP — PR #208 at `72eeef802813f20de7d039fca74d995b79773548`, base `fc831639bba42946ebcd2e7b8fac01ddde7b6c41` (codex independent review). No blocking findings.

- Exact three-file scope verified. All four deadline-set failure sites use the same classifier and stage as their neighboring transport failures (`internal/hubclient/session.go:144,158,324,353`). Removing the two explicit hello-path closes is correct: `classifySSHFailure` closes the transport itself (`internal/hubclient/ssh.go:352`). Its classification arms are unchanged.
- Classification is the right fix. Moving the read deadline before the write only changes the timing of one failure; it leaves other deadline errors without the established transport-error classification.
- `transmitted` remains false before a failed write deadline and true after a completed request write. A Send whose read deadline fails after transmission now reaches the existing `E_SEND_UNKNOWN` conversion. No output has been emitted at these sites, so they do not become `E_RESULT_UNKNOWN`; that conversion remains inside the streaming loop.
- The deterministic transport fixtures force closure on either side of the read-deadline call and cover all four call sites (`internal/hubclient/deadline_lost_test.go`). Comparing hello failures with their neighboring read-path classification is adequate for this scoped equivalence claim; the classifier was independently inspected and currently yields `E_CONNECT` for these fixtures.
- Independent validation PASS with all AGENTCHUTE_* variables stripped: `go test -race ./internal/hubclient -count=1`. Current CI has six successful jobs and one skipped upgrade rehearsal. Coordinator supplied the separate full-suite gate; this review used targeted tests per E5.

